PR #2, opened 2022-03-30
Add Moodle App guide

NoelDeMartin:
I've migrated some of the Moodle App documentation, here's some things I noticed:

- I never use `yarn` so I installed it with `npm` anyways, it worked well.
- We're missing a `.nvmrc` file, I was using node v14 and I got an error because `String.replaceAll` is not supported.
- At first it was confusing to create new pages. I started creating them in `/docs`, but that didn't work because the links at the top take you to the 4.0 docs. Later I noticed that clicking the "quick start" link on the home page takes you to the next version. So it is confusing because depending where you click, it takes you to a different version. Although I guess this is fine once you get used to it, but it could trip up some contributors getting started.
- It would be nice to have something to automatically rename all references when files are renamed. Whilst I like the pure markdown approach, maybe we could have some scripts in CI checking for link integrity (at least internal links).
- I noticed that some links are written using a convention for writting them at the bottom of the file instead of inline markdown. If these such practices are mandatory, I think we should also do something in CI or somewhere else to enforce it because it's very easy to forget (I haven't done it at all in this PR).

Other than these small things, I think it was quite easy to get started. I haven't done anything complex, but it seems like it would be easy to learn more.

### DIFF
--- a/docs/guides/moodleapp/01-overview.md
+++ b/docs/guides/moodleapp/01-overview.md
@@ -1,0 +1,62 @@
+---
+title: Moodle App Overview
+---
+
+The Moodle App is a mobile application that helps users make the best of their Moodle sites on handheld devices. It has some additional features like offline access, and a dedicated interface adapted to mobile. It's focused on student functionality, so you won’t find all the features you have on the web for teachers and admins. You can learn more about the features available in [the user documentation](https://docs.moodle.org/).
+
+On a technical level, it's a completely different codebase from the Moodle LMS, and interacts with a Moodle site using [web services](#). You can find the source code of the application in github: [github.com/moodlehq/moodleapp](https://github.com/moodlehq/moodleapp).
+
+Before embarking into any Moodle-specific documentation, we recommend that you are at least familiar with [Angular](https://angular.io/) and [Ionic Framework](https://ionicframework.com/). These are the core technologies used in the application. We'll reference any relevant concepts, but having a basic idea will take you a long way in understanding the Moodle App.
+
+## Basic concepts
+
+### How does it work?
+
+When you are accessing a Moodle site on the web, you are only capable of using one site at a time. In contrast to that, the Moodle App can be used with multiple sites at once. However, you will need to switch sessions to interact with each site, so you won’t be able to use multiple sites simultaneously (but you'll continue receiving push notifications and reminders for all the sites connected in the app).
+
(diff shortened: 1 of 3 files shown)


andrewnicols:
Thanks Noel,

> I never use yarn so I installed it with npm anyways, it worked well.

I've not used Yarn either but it seems a little bit nicer to user than npm. Either will work but they recommend not mixing the two

> We're missing a .nvmrc file, I was using node v14 and I got an error because String.replaceAll is not supported.
Thanks - we should fix that

> At first it was confusing to create new pages. I started creating them in /docs, but that didn't work because the links at the top take you to the 4.0 docs. Later I noticed that clicking the "quick start" link on the home page takes you to the next version. So it is confusing because depending where you click, it takes you to a different version. Although I guess this is fine once you get used to it, but it could trip up some contributors getting started.

Yeah - that's something to address. I've versioned the docs at the moment to see how versioned docs will work for us.
There's a bit of configuration still to sort out with it once we work out how we want to do things.

> It would be nice to have something to automatically rename all references when files are renamed. Whilst I like the pure markdown approach, maybe we could have some scripts in CI checking for link integrity (at least internal links).

Yes - I agree about this.
I've actually created a remark plugin which ensures that _internal_ links within the tracker continue to work. That is to say that a link like `[[Coding_style|Moodle coding style]]` will generate a link to `https://docs.moodle.org/dev/Coding_style` with the title `Moodle coding style`.
Unfortunately having the external links point to other things automatically gets a little messier because the syntax is so generic.

> I noticed that some links are written using a convention for writting them at the bottom of the file instead of inline markdown. If these such practices are mandatory, I think we should also do something in CI or somewhere else to enforce it because it's very easy to forget (I haven't done it at all in this PR).

When I was playing with the Sphinx approach I originally used RST which mandates that method. I haven't gotten around to updating those for the Markdown but I think that we should as they're weird. The main benefit is that you can reduce the number of places to maintain references, but they're a little non-standard.

I've adjusted your files a little to rename the files and instead handle the sidebar order with the `sidebar_position` front matter as described at https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#markdown-front-matter.

Since the Mobile app is versioned and managed slightly differently to Moodle itself, we may want to consider how its documentation sits too - there are a number of options to consider including having it in it's own separate un-versioned area.

I've just pushed a number of other changes to the docs, including an update to the "Community" section, which is not versioned. This includes things like the Roadmap, policies, Coding style, Developer meetings, and a few other areas. Those are also an unversioned area with their own sidebar and navigation.

PR #6, opened 2022-04-07
Add Tracker introduction

sarjona:
I've been playing with the dinodevdocs repository, to see how it works. I've created the page for https://docs.moodle.org/dev/Tracker_introduction as a PoC. Some comments about this PR:
- It was easy to create it. The "most difficult" part was to decide where to place it (so I'm open to suggestions).
- I love the :::important/tip/note... It helps to identify them easily on the page.
- I edited the community.js to display "Tracker introduction" in the left menu.
- While creating the page, I've updated some parts and also removed a section that I think it's not required (the less, the better).


andrewnicols:
Hi Sara,

Have a look at https://github.com/andrewnicols/dinodevdocs/compare/trackerintro

I've modified the structure to move all of those docs under Process.

If you try running `yarn build` on your branch, you'll see that it fails because there are some internal links which are orphaned by this change which fails the build.

See what you think :)

PR #8, opened 2022-04-08
Restructure the community section

andrewnicols:
Restructure the community section

    The community section included resources for both the community, and
    development standards. These are both unversioned items, which do not
    belong in the version-specific documentation.

    To better organise these, I've renamed the 'community' to 'general', and
    then organised the existing documentation within that section.

    I've generated separate sidebars for the general folder, and linked to
    the initial content for each via the navbar.

    All existing links are updated, and a couple of new entrypoints have
    been introduced.

    I've also moved the Develop Meetings into /general/community/meetings.

    At this point I have _not_ renamed the meetings to 'Developer meetings'
    or similar. That should probably be a separate PR once a decision is
    made on that content.


sarjona (approved):
Hi Andrew!
Thanks for restructuring the community section. I like the new organisation! 

About the "Meetings" vs "Developer meetings", although this is dev docs, I think it's a good idea to rename it (it will help to understand better what "Meetings" is).

And one more thing (unrelated to the patch): I keep thinking "Moodle App Coding Style" should be placed in "Process&policies > Policies" (and then, we can use the "See also" section to link both docs (which will be placed in different categories).

PR #9, opened 2022-04-08
Add Peer reviewing

sarjona:
As this page has been added to the "Process" category, the process/index file has been moved outside this folder.
Apart from that, this commit also updates references to the Peer reviewing page (there were a couple of them in process.md).


andrewnicols:
Hi Sara,

I'd recommend a slightly different structure:
* Don't rename the process.md
* Create a new directory for the files in it - I'd recommend `_process`. Directories with an underscore are not included  in any searches

That should be sufficient and will prevent the need to rename the file everywhere that it's included (and thus fix the build issues).

andrewnicols:
I took another look at this today and decided that it would be better to just move the resources into `process/_files/` and fix the broken internal links highlighted by the `yarn build`.

I've made those changes and merged the branch.

sarjona:
+1 to add the resources into the __files_ folder! I forgot again to run yarn build... grrrrr! Thanks for fixing the errors, Andrew!

PR #10, opened 2022-04-08
Add linting and style checks

andrewnicols:
These commits add:
- linting using `markdownlint-cli`
- improvements to README.md to document this change and how to run it
- fixes for existing linting issues

We can refine these rules over time, but I feel that it's better to start strict and become less so, than to start our journey without many rules and then need to add them and update an increasing amount of documentation in the process. These rules really aren't too contentious anyway.

### DIFF
--- a/.env.default
+++ b/.env.default
@@ -1,0 +1,18 @@
+# URL
+# ===
+# The default URL for build content is https://develop.moodle.org.
+# For local development you may wish to specify an alternative for your own environment.
+#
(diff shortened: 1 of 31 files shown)


sarjona (approved):
The markdownlint tool is great and working as expected. Besides, I agree with you that it's better to add these rules at this stage of the project (it will help us to get some consistency... and consistency is always welcome!).
I'm happy to approve and merge this patch! Well done!

PR #14, opened 2022-04-09
Stop showing the last update time/date/author

andrewnicols:
These currently seem to be the last time that any change was made to the
repository, not just the file in question, which makes it somewhat
meaningless for us.


andrewnicols:
I'm reconsidering this one. When generated locally the dates are correct so I wonder if it may be because I force-pushed the last commit to get the build working properly.

andrewnicols:
I'm going to submit this for review.

I suspect that this does not work where we mereg commits using a merge request. Docusaurus project don't perform merges, which preserves authorship properly I think.

andrewnicols:
Looking again, it looks like they're all correct after all. They're just messed up from some lint fixes.

PR #20, opened 2022-04-11
Move /docs/apiguides to /docs/apis

andrewnicols:
This is just clearer. These are guides to the APIs but they would be
clearer under the `api` folder instead.


sarjona (approved):
Thanks for working on this. I agree with you that _apis_ is clearer than _apiguides_ so I'm happy to approve it :-)

PR #33, opened 2022-04-11
[repo] Rename about to description

andrewnicols:
This commit fixes all of the issues with the yml files now we've found how to validate them /o\


sarjona (approved):
Yes, thanks for fixing them! Now they should work (because, at least, they are displayed properly when accessing to these files from the repository!).

PR #40, opened 2022-04-12
Add pre-commit tests

andrewnicols:
This change adds a few bits all in one:
* eslint configuration
* cspell configuration
* husky and lint-staged to perform a pre-commit hook lint rewrite and spell-check

This is not quite ready, but it would be good to try it otu and see what you think.

<a href="https://gitpod.io/#https://github.com/andrewnicols/dinodevdocs/pull/40"></a>


sarjona:
Thanks for keep adding interesting and useful tools to the devdocs repository!! 

I've been playing with your branch and the husky and lint-staged tools were working as expected (I've also seen they have helped to identify and fix some styling issues in JS... Well done!).

However, while playing, I missed the option to run eslint without having to add these files to stash (besides, while fixing the errors, I forgot to add the file again to stash). 
Am I doing something wrong/missing anything or is that expected? I can see it will help to identify issues when we're pushing changes (as CiBoT does), but I also feel it would be great to be able to run them locally while you're working on a patch (before committing changes).

PR #50, opened 2022-04-15
[repo] Add Contributions guide

andrewnicols:
This is the start of a contributions guide.


andrewnicols:
I'm not sure of the best way to link this into the nav.

I'm worried that we're going to end up with too mcuhin the top navbar for example.

For the moment I've put it in the footer.

PR #51, opened 2022-04-16
Introduce pre-commit hooks to lint and check more content

andrewnicols:
This issue adds:

- lintstaged
- husky
- eslint
- cspell

The [lintstaged](https://github.com/okonet/lint-staged) tool lints all staged files through the configured linters
[Husky](https://typicode.github.io/husky/#/) is a NodeJS package which sets up some awesome integrations with Git hooks.

We can use these together to create a pre-commit hook to lint all files before commit.

I've pulled an eslint similar to the one used by upstream docusaurus which is a combination of the airbnb style guide, and react guides. One of the features is that it adds a copyright header to any JS file.


sarjona (approved):
Thanks a lot for working on this, Andrew. I especially love the Spell-checker (and now that I've been playing a little bit with husky, I can also see how useful it will be!!).

I think there are still a few files with errors but they can be fixed in the future (for instance `sidebars/docs.js:28:18 - Unknown word (gettingstarted)` )... Let's get this patch merged and then fix them in a separate issue! :-)

PR #56, opened 2022-04-17
[repo] Fixes for migratedPages

andrewnicols:
There are two commits here:

- style fix for `migratedPages.js` to set array keys as unquoted strings; and
- bug fix when suggesting a page name


andrewnicols:
Self-merging as this affects several other PRs

PR #57, opened 2022-04-19
[docs] tools/phpcs: Document new phpcs config file

andrewnicols:
This file will be added by MDL-74511.

This issue should not land until MDL-74511 lands.


andrewnicols:
An updated version of this PR have been pushed. I've removed a lot of unneceessary information and restructured a little too now that we have both a `.phpcs.xml` and `.phpcs.xml.dist` file.

I've also updated these docs to provide a nicer way to set a per-default configuration which is automatically read by editors and the CLI tool.

PR #59, opened 2022-04-19
[docs] api/files: Migrate file API docs

andrewnicols:
Migrated from:
https://docs.moodle.org/dev/File_API
https://docs.moodle.org/dev/File_API_internals

### DIFF
--- a/docs/apis.md
+++ b/docs/apis.md
@@ -20,7 +20,7 @@ The [[Data manipulation API]] allows you to read/write to databases in a consist
 
 ### File API (files)
 
-The [[File API]] controls the storage of files in connection to various plugins.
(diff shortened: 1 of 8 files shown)


sarjona (changes requested):
These pages look great: I like the way they are organised and their content. 
As you'll see, there is only one minor thing to review for merging it! :-)

andrewnicols:
Thanks Sara,

Good catch on the tracker link.

I've addressed some additional markdown linting, and spelling issues that I hadn't spotted eralier too.

sarjona (approved):
Hi Andrew! Thanks for reviewing the patch and addressing a few more things... Well done!! :-)

PR #63, opened 2022-04-20
[repo] Create a basic fetch script

andrewnicols:
This is a really basic system to fetch a page from the legacy docs into
a new location on disk, and perform basic changes.

It does *NOT* catch everything and it can do better, but it's what I've
been doing manually.

At the moment it will:

- not convert any table
- complain about migrated pages but not fix them
- do nothing with tracker links
- not update the migratedDocs.yml file with it's existence

I've tried to merge some of the functions and utilities that I
previously wrote for the legacy wiki updater into a shared utils file
too.

Usage:

```
node scripts/fetchLegacyPage.js migrate [Legacy_Page_Title] [path/to/new/page.md]
```

### DIFF
--- a/scripts/wikimedia-fetch.js
+++ b/scripts/wikimedia-fetch.js
@@ -1,0 +1,168 @@
+#!/usr/bin/env node
+/**
+ * Copyright (c) Moodle Pty Ltd.
+ *
+ * Moodle is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Moodle is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


andrewnicols:
Rewritten to:

```
node scripts/wikimedia-fetch.js migrate Authentication_plugins docs/apis/plugintypes/auth.md
```

scripts/wikimedia-fetch.js
```diff
@@ -0,0 +1,168 @@
...
+
+const remoteHost = 'docs.moodle.org/dev';
+
+const client = getClient(remoteHost);
+const logger = getLogger();
+const fetchDoc = getFetchDoc(logger)(client);
+
+const getFrontpage = (title, content) => {
+    const data = {
+        title: title.replace('_', ' '),
```
sarjona, line 48: replaceAll should be used (because sometimes titles have more than one '_'

sarjona (changes requested):

sarjona (approved):

PR #65, opened 2022-04-21
[repo] The update-wikimedia script should only run on our repo

andrewnicols:


sarjona (approved):
Thanks for fixing this, Andrew!! :-*

PR #68, opened 2022-04-21
[repo] Correct expression to only run on moodle/devdocs

andrewnicols:
Before it only ran if we pushed directly to main without a pull request.

We should be comparing the `github.repository`, documented at
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
to the canonical repository.


andrewnicols:
Self-merging this to prevent further issues in the event someone merges anything else first.

PR #69, opened 2022-04-22
[docs] Migrate the Plugin_files page

andrewnicols:


andrewnicols:
This issue needs a bit of explaining.

As I was moving some of the plugintype doumentation over I noticed that there is a lot of documentation relating to plugin files which is duplicated, or out-of-date. We often document files like `version.php` in plugintype docs, but each is different, with different levels of detail, etc.

To solve this I thought it would be useful to move these sections out into reusable templates, but then I realised that we sometimes want to specify extra information, differnt information, different examples, no examples, etc.

To solve this I created a `ComponentFileSummary` component which takes the key information required to generate a summary of a file, that is information including:

- plugintype
- pluginname
- filepath
- brief summary of purpose
- example
- description

We can use this data to create a consistent look and feel which is reusable.

The default description of the file is provided in an mdx file. Additional descripton data can be added using the `extraDescription` property, or entirely replaced with the `description` property.

Likewise a default example can be provided, but this can be modified from a specific plugintype.

Equally the `plugintype`can be specified in a specific component type (e.g. antivirus) and the title showing the path to the file will be updated using data from the standad Mooodle components.json. For example, the common files shows information for a generic `version.php`, but the activity module docs adds its own using:
```
<VersionPHP
    plugintype='mod'
    pluginname='[modname]'
/>
```

Likewise the Antivirus documentaiton wishes to specify a better example for its settings.php.  This can be achieved with:
```
export const settingsExample = `
if ($ADMIN->fulltree) {
    $settings->add(
        new admin_setting_configexecutable(
            'antivirus_scanmyfile/pathtoscanner',",
            new lang_string('pathtoscanner', 'antivirus_scanmyfile'),",
            new lang_string('configpathtoscanner', 'antivirus_scanmyfile'),",
            ''",
        )
    );
}
`;

<SettingsPHP
    plugintype="antivirus"
    pluginname="scanmyfile"
    example={settingsExample}
/>
```

The standard description will beu sed, but the example filename will have the path `lib/antivirus/scanmyfile/settings.php`. The `lib/antivirus` comes from the standard Moodle `components.json` and the `plugintype="antivirus"`. The `scanmyfile` comes from `pluginname`, and the `settings.php` is specifeid in the generic antivirus settings as the default filename to use. The custom example is specified as an **exported constant** and passed into `example`.

docs/apis/_files/locallib.mdx
```diff
@@ -0,0 +1,11 @@
+:::caution Legacy feature
```
sarjona, line 1: I love to see it has been marked as a legacy (I think that's a good idea). While reviewing the whole patch, I was wondering why the Common files (and also the xxx Plugins) are versioned. Wouldn't be better to have them unversioned and then add some badge to mark things version-dependant? 
I guess I'm missing some scenario where will be useful to have this information versioned (that's why I'm asking about it).

docs/apis/_files/db-install-xml.mdx
```diff
@@ -0,0 +1,7 @@
+The `install.xml` file is used to define any database tables, fields, indexes, and keys, which should be created for a plugin during its initial insatllation.
+
+:::note
```
sarjona, line 3:
```suggestion
:::caution
```

sarjona: I think that's quite important, so it will be better to mark using some other admonition, like caution.

docs/apis/_files/db-access-php.tsx
```diff
@@ -0,0 +1,56 @@
...
+            'editingteacher' => CAP_ALLOW,
+        ],
+    ],
+];
+`;
+
+export default (initialProps: Props): PluginFileSummary => {
+    const props = fillDefaultProps({
+        filepath: '/db/access.php',
+        summary: 'Plugin capabilities',
```
sarjona:
```suggestion
        summary: 'Plugin capabilities',
        refreshedDuringUpgrade: true,
```

sarjona: Shouldn't the Upgradable badge be displayed for db/access.php too? Capabilities are also updated when the upgrade is performed...

docs/apis/_files/pix-dir.mdx
```diff
@@ -0,0 +1,5 @@
+Plugins can provide icons in several formats, and most plugin types require that a default icon be provided.
+
+Where a browser supports it, the `svg` format is used, falling back to `png` formats when an SVG is unavailable.
+
+Full details of the correct naming, sizing, and design guidelines for icons in Moodle can be found in the [[Moodle icons]] documentation.
```
sarjona, line 5: The https://docs.moodle.org/dev/Moodle_icons is pointing to a project page which, from 4.0 onwards, is kind of deprecated... So probably we should create a separate page for that at any point (it's out of the scope of this issue, so maybe it worth creating a separate issue to not forget about it).

docs/apis/_files/version-php.mdx
```diff
@@ -0,0 +1,10 @@
+The version.php contains metadata about the plugin.
```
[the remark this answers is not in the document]
sarjona: I've just seen the content in https://docs.moodle.org/dev/version.php seems to be migrated in the file version.php included in this commit ( https://github.com/moodle/devdocs/pull/69/commits/e0b43b7c0229dd2a6633b0d77f105503ea7ac33d#diff-0dc418129946563e59bb69680a5ef0ded4329ed9e56e3d295c25e12ea726726b ) but I can't see where/how it's displayed...

version.php
```diff
@@ -0,0 +1,145 @@
+**Note:** Support for <tt>$plugin->supported =</tt> and <tt>$plugin->incompatible =</tt> was added in Moodle 3.9 (MDL-59562).
```
sarjona, line 1: Where is all this content file displayed?

docs/apis/plugintypes/qbank/index.mdx
```diff
@@ -0,0 +1,25 @@
+---
```
sarjona, line 1: As this page is incomplete, is there any issue to add the missing information?  For now, I think it's OK to leave it as it is, but it would be great to mark it as "WIP" to add more information at any point in the near future :-)

docs/apis/plugintypes/local/index.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+## Examples
+
+The following examples show some ways in which you can use a local plugin.
+
+### Adding an element to the settings menu
+
+A local plugin can extend or modify the settings navigation by defining a function named `local_[pluginname]_extend_settings_navigation` in its `lib.php`. This is called when Moodle builds the settings block. For example:
+
+export const settingsNavigationExample = `
```
sarjona, line 51: This information is not displayed on the page, nor any of the examples.

sarjona (changes requested):
Thanks a lot for working on this huge feature to simplify the creation of the plugin type pages! I think that's a great idea and, although the visualization can still be slightly improved, it's a very good improvement! Well done.

As you'll see, I've raised a few questions and sent some suggestions. Apart from that, I have one more comment, related to the way categories are displayed: as you can see in the following screenshot, content below each plugin type is not consistent (for instance, a couple of them are displaying the <Since xxx label):
plugintypes

version.php
andrewnicols: Ah - this is a mistake and needs removing.

docs/apis/plugintypes/qbank/index.mdx
andrewnicols: This is the current state of the page in the current dev docs.

We currently dont' have anything to do that, but I would like to add something to the frontmatter, like `draft: true` which we then display as a banner and ask for help in fixing.

docs/apis/_files/pix-dir.mdx
andrewnicols: Yeah - I agree.  I've created #82

docs/apis/_files/locallib.mdx
andrewnicols: I feel that this stuff is specific to Moodle development as a whole and probabyls houdl be versioned.

Right now the docs are split into:

- docs - stuff related to Moodle development which is for a version of Moodle
- general - stuff which does not need a version

However, there's also an additional divide:

- docs: Only contains development stuff
- general: Stuff which is meta to development - like Process, tooling, etc.

Ultimately there is no _harm_ to having docs in the versioned wiki where it's about development, but if we move it into general then it does becomes slightlyt harder to link between things. For example, the page tags are within a specific instance of the docs (docs, and general are different instances of the the docusaurus-plugin-docs plugin).

So tl;dr: My feeling is that these _should_ be documented. We can add things like `<Since>` tags, and legacy, deprecated, etc. and over time those will be removed when they're removed from Moodle.

docs/apis/_files/lib.mdx
```diff
@@ -0,0 +1,9 @@
+The `lib.php` file is a legacy file which acts as a bridge between Moodle core, and the plugin. In recent plugins it is should only used to define callbacks and related functionality which currenltly is not supported as an autoloadable class.
```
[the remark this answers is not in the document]
andrewnicols: Ideally we should have a subpage of the commonfiles with more information for lib.php which we link to, but I think that should be a separate migration. This one is already hard enough lol

docs/apis/_files/version-php.mdx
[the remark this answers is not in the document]
andrewnicols: It's not.  It was mistakenly included /o\

I'll aim to move this to a subdir of commonfiles in a separate issue. This one is too big already.

docs/apis/_files/db-access-php.tsx
andrewnicols: Nope - they aren't refreshed!! New ones are _added_ but existing ones are not modified.

docs/apis/plugintypes/qbank/index.mdx
andrewnicols: I've raised #83 to do something like this.

I'm happy to remove this doc from this PR for now instead.

andrewnicols:
> As you'll see, I've raised a few questions and sent some suggestions.

I think I've resolved all but one of these appropriately. The final one is awaiting feedback from you re the question bank plugin type.

> Apart from that, I have one more comment, related to the way categories are displayed: as you can see in the following screenshot, content below each plugin type is not consistent (for instance, a couple of them are displaying the <Since xxx label):

This is a bug in Docusaurus. I've asked on their Discord chat if there's an existing issue for it - waiting to hear back.

I'll rebase my code now (some conflicts to resolve) and then I'll update the patch.

docs/apis/plugintypes/local/index.mdx
andrewnicols: http://localhost:3000/devdocs/docs/apis/plugintypes/local#adding-an-element-to-the-settings-menu

docs/apis/_files/lib.mdx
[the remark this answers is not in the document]
sarjona: Thanks for the clarification. I agree it can be done in a separate issue (I just wanted to know how these pages fit here) :-)

docs/apis/plugintypes/local/index.mdx
sarjona: Now it's working as expected :-)

docs/apis/plugintypes/qbank/index.mdx
sarjona: As you've raised #83, I think it's OK to leave it as it is now and then add the `draft` setting, tag or both.

sarjona (approved):
Hi Andrew!
Thanks a lot for clarifying the raised points :-)
The patch looks good and I think it's ready to be merged. Besides, I've just seen the issue with the categories has been fixed too! <3
plugintypesfixed

PR #71, opened 2022-04-25
[docs] process/triage: Migrate Bug triage

sarjona:
This patch migrates the https://docs.moodle.org/dev/Bug_triage page
I've removed the triage image on that page because I don't think it's useful having it there.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/71"></a>


general/development/process/triage.md
```diff
@@ -0,0 +1,233 @@
...
+
+Tracker is set up by default so as soon as you comment on any issue or edit it, you become an automatic watcher and any following change on the issue will be emailed to you.
+
+If you have encouraged the reporter well, they may **submit a patch** or somebody else may do it. Make sure that the `patch` label is added or issue is sent to `Peer review`. On the contrary, if `patch` label was added by somebody else but you clearly see that patch is far from ready, remove the label and leave a comment explaining it.
+
+Users may also comment with additional details, screenshots, replicating instructions. It may happen that the issue gets re-evaluated and priority or summary changed.
+
+### Revisiting old issues
+
+While searching the tracker you may come over issues that were reported long time ago but still remain open. Again, ask the following questions:
```
andrewnicols, line 91:
```suggestion
While searching the tracker you may come over issues that were reported a long time ago but still remain open. Again, ask the following questions:
```

general/development/process/triage.md
```diff
@@ -0,0 +1,233 @@
...
+- **Is this still an issue?** If it is not applicable any more either close it or comment about it on the issue and recommend to close. Very few tracker users actually have permission to close issues but the component lead or watchers will see your comment and revisit issue. Another good practice is to replace `triaged` label with `triaging_in_progress` and add a comment asking if the issue can be closed.
+
+:::warning Closing issues
+
+If users confirm that it was resolved or nobody replies in 30 days, issue should be closed.
+
+:::
+
+- **Do affected versions need correction?** If the issue is still applicable, make sure to add missing current affected versions or comment about it on the issue if you can't edit it.
+- **Does the issue have patch and if yes, is it still applicable?** If the issue has patch that still works but `patch` label is missing, look through comments or history to see if the `patch` label was removed after reviewing the current patch. If you find that label was never added, do it yourself. On contrary if the issue has `patch` label but the patch is no longer applicable or not sufficient, remove the label.
```
andrewnicols, line 102:
```suggestion
- **Does the issue have a patch and if yes, is it still applicable?** If the issue has a patch that still works but the 
`patch` label is missing, look through comments or history to see if the `patch` label was removed after reviewing the current patch. If you find that label was never added, do it yourself. If the issue has the `patch` label but the patch is no longer applicable or not sufficient, remove the label and comment explaining why.
```

general/development/process/triage.md
```diff
@@ -0,0 +1,233 @@
...
+- **Are there any duplicating issues?** When finding duplicates among the old issues it might not be obvious which issue to close as a duplicate. Usually we should leave the first reported issue but if the later issues have more watchers, better description, more votes, useful comments, attached patch, etc. you may decide to close the earlier issue and leave the later. Sometimes both issues have lots of watchers and they both remain open.
+
+:::tip
+
+In any case, always create links between duplicates or related issues.
```
andrewnicols, line 107:
```suggestion
Always create links between duplicates or related issues, whether you close an issue or not.
```

general/development/process/triage.md
```diff
@@ -0,0 +1,233 @@
...
+
+:::
+
+- **Does the issue have assignee who forgot about it or misleading status `Development in process`?** Due to some process changes in 2013 some issues still have a real user in the **Assignee** field but this user actually does not work on the issue. Sometimes **Assignee** remains filled after failing `Peer review`, sometimes developers simply forget that the issue was assigned to them. If you suspect that **Assignee** does not actually work on the issue, comment asking they about it and in some cases remove the **Assignee** completely. Allow somebody else to work on the issue without feeling that the issue is "taken". Please also note that for some time tracker had a restriction that **Assignee** could not be empty so you can find lots of issues assigned to **moodle.com** or **Nobody**. Do not modify such issues as this creates unnecessary activity, emails to watchers and irrelevant change in the "Last update date".
```
andrewnicols, line 111:
```suggestion
- **Does the issue have an assignee who forgot about it or a misleading status (for example `Development in progress` for a long time)?** Due to some process changes in 2013 some issues still have a real user in the **Assignee** field but this user does not actually work on the issue. Sometimes the **Assignee** remains filled after failing `Peer review`, sometimes developers simply forget that the issue was assigned to them. If you suspect that the **Assignee** is not actually working on the issue, comment asking they about it and in some cases remove the **Assignee** completely. Allow somebody else to work on the issue without feeling that the issue is "taken". Please also note that for some time the tracker had a restriction that **Assignee** could not be empty so you can find lots of issues assigned to **moodle.com** or **Nobody**. Do not modify such issues as this creates unnecessary activity, emails to watchers and irrelevant change in the "Last update date".
```

general/development/process/triage.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Comments templates
+
```
andrewnicols, line 114: I'd suggest adding a tip admonition to tell people to check out the Canned responses.

I'd almost be tempted to suggest we entirely remove these and only use Canned responses beacuse these will likely become stale and out-of-date over time.

general/development/process/triage.md
```diff
@@ -0,0 +1,233 @@
...
+If you can propose a code solution, that will help others who may have the same need and will increase the chance of this improvement/feature coming about sooner. If you are able to provide a patch or links to your Git repository branch, please add a patch label so we will spot it. Otherwise, how about posting in a forum on moodle.org and encouraging people to vote, comment and/or come up with a patch.
+
+For more ways to maximize the chance of your idea being implemented, see the guide [new feature ideas|https://docs.moodle.org/dev/New_feature_ideas].
+```
+
+## Creating a filter and gadget for triaging
+
+If you are a component lead you are responsible for triaging issues that involve your component. A good way to monitor newly created issues is to create a filter in the Tracker and add a gadget on your dashboard to show the results of the filter.
+
+### Adding a filter
```
andrewnicols, line 189: Maybe we should move these to the https://moodle.github.io/devdocs/general/development/tracker/tips page  and link there instead? Maybe not. Not sure.

andrewnicols (changes requested):
These are mostly just grammatical fixes. There are a few which suggest some other changes like removing an entire section, moving a section to another doc, and adding an admontion.

andrewnicols (approved):

general/development/process/triage.md
sarjona: Hi Andrew! 
I had the same thought while reviewing this page so your comment has convinced me to remove these template messages from this page and replace them with a reference to canned responses in the tracker :-)

general/development/process/triage.md
sarjona: You're right! I think these sections (Creating a filter and Adding a filter in a gadget to your dashboard) fit better in the "Tracker Tips&Tricks" page so I've moved these changes there, improving the pre-existing information and adding a reference from the Bug triage page ;-)

PR #87, opened 2022-04-28
[docs] community/credits: Migrate Credits docs

sarjona:
I've migrated the Credits pages into the Community. However, I'm not sure
if that's the best place for the Third-party libraries (because it's kind
of Moodle-version related). So maybe it's better to place it in the docs
and then link it from there.
Thoughts?

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/87"></a>


andrewnicols:
Hi Sara,

My thinking with the libraries is that we should:

- update the existing thirdpartylibs.xml files in core and add:
  - whether customisations have been made (`<customised />` if true)
  - the URL source of the library
  - the copyright owner
- create a script to parse the various thirdpartylibs files and turn them into markdown
- turn that into a Github action run weekly

I'd say it's fine to include it in the patch for now, but let's do something better in the long run.

andrewnicols (approved):

sarjona:
Hi @andrewnicols!
Thanks a lot for reviewing the patch:
- I've created https://github.com/moodle/devdocs/issues/93 to deal with the libraries (I was thinking about that while reviewing the page but, due to the time restrictions and taking into account it was a single page, I decided to leave it as it was for now). But I agree that's something we could automatise and improve in the future. 
- I've also raised https://github.com/moodle/devdocs/issues/94 to automate the generation of the docs and dev docs contributors :-) I've created a single issue but we could split it in the future if required ;-)

PR #88, opened 2022-04-28
Some typography considerations.

lucaboesch:
I've been skimming through the new dev docs and I like them a lot.
I noticed that the typography is used differently and it would be maybe a good thing to use the ’ or ' consistently.
But I'm maybe too picky.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/88"></a>

### DIFF
--- a/docs/guides/javascript/index.md
+++ b/docs/guides/javascript/index.md
@@ -325,8 +325,8 @@ A lot of this content is being migrated to use Mustache Templates which are the 
 
 Where content is generated in PHP you will need to include your JavaScript at the same time.
 
-Although several older ways to include JavaScript from PHP, it`s strongly
-recommended that all new JavaScript only use the`js_call_amd` function on the
+Although several older ways to include JavaScript from PHP, it's strongly
+recommended that all new JavaScript only use the `js_call_amd` function on the
 `page_requirements_manager`.
 This has a similar format to the version used in Templates:
 
@@ -371,7 +371,7 @@ export const init = ({courseid, category}) => {
 
 :::caution
 A limit applies to the length of the parameters passed in the third argument.
-If data is already available elsewhere in the DOM, you shoudl avoid passing it as a parameter.
+If data is already available elsewhere in the DOM, you should avoid passing it as a parameter.
 :::
 
 ## Passing data to your Module
--- a/docs/guides/moodleapp/customisation/custom-apps.md
+++ b/docs/guides/moodleapp/customisation/custom-apps.md
@@ -9,5 +9,5 @@ Unfortunately, the core team does not have the resources to help people use the 
 
 [MoodleHQ](https://moodle.com/) provides a service where they can customise the app for anyone who needs it including the Push Notifications service management with Airnotifier. This service also helps fund the overall development of the app: [moodle.com/branded-app/](https://moodle.com/branded-app/).
 
-If you’re still considering making your own custom app, you can read the rest of the
-documentation in this wiki to understand how the application works and you’re also welcome to ask for help in [the forum](https://moodle.org/mod/forum/view.php?id=7798) or [[Moodle Mobile custom apps|read the legacy documentation]] (no longer maintained).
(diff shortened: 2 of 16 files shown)


sarjona (changes requested):
Hi @lucaboesch!
Thanks a lot for looking at the new devdocs and raising this.
I think it's better to leave \' instead of replacing it with \` .
However, I've seen you' fixed a couple of things too (shoudl and it's -> its). So could you please remove the changes done with ` and leave only these in order to merge the patch? TIA! :-)

lucaboesch:
Good morning @sarjona.
I see where you're getting at, so it would be of personal taste of the editor, with one person writing it is ` and with another it is '
We could also have the ` changed into ' …
What do you think?

andrewnicols:
Personally I'm not keen on this change. We should use the single (`'`) for contractions (there's, it's, doesn't, etc.) because that is the key on the keyboard that one types. A standard English (both US and UK variants) has no apostrophe character (`’`) - to get one Mac OS one must use `Option + Shift + ]`.

The only options availble to us are:
- Backtick `
- Quote `"`
- Single quote `'`
- Apostrophe `’`

We _CANNOT_ use the backtick as that is used for code. We _SHOULD NOT_ use the double-quote, because that is jsut wrong.
Although the Apostrophe is probably more typographically correct, it's also extremely non-obvious for a vast majority of our devs and does not appear on standard English keyboards.

lucaboesch:
Probably I've not gotten each and every \’ but I've switched those to '.

docs/guides/javascript/index.md
```diff
@@ -325,9 +325,9 @@ A lot of this content is being migrated to use Mustache Templates which are the
 
 Where content is generated in PHP you will need to include your JavaScript at the same time.
 
-Although several older ways to include JavaScript from PHP, it`s strongly
-recommended that all new JavaScript only use the`js_call_amd` function on the
-`page_requirements_manager`.
+Although several older ways to include JavaScript from PHP, it's strongly
+recommended that all new JavaScript only use the \`js_call_amd\` function on the
```
sarjona, line 329:
```suggestion
recommended that all new JavaScript only use the `js_call_amd` function on the
```

docs/guides/javascript/index.md
```diff
@@ -325,9 +325,9 @@ A lot of this content is being migrated to use Mustache Templates which are the
...
+\`page_requirements_manager\`.
```
sarjona, line 330:
```suggestion
`page_requirements_manager`.
```

sarjona (changes requested):
Thanks Luca for making these apostrophes more consistent!! As you'll see, there are only a couple of places where Backtick shouldn't be escaped.

sarjona (changes requested):
Hi Luca!
Thanks, the patch looks great! I have only comments about git: it would be great if you could squash your commits in 1 (I can't see the value of having separate commits for these changes). And it would also be amazing if you could reword it, adding "[docs]" as prefix (you can look at the history of this repository to see what I mean).
This is still a WIP and we are working on a page to explain how to contribute to this devdocs (with things like this to consider) :-)
Tons of kisses for you, Luca! :-*

lucaboesch:
Done, @sarjona and thousands of kisses back!

docs/guides/javascript/index.md
sarjona, line 329:
```suggestion
recommended that all new JavaScript only use the `js_call_amd` function on the
```

docs/guides/javascript/index.md
sarjona, line 330:
```suggestion
`page_requirements_manager`.
```

sarjona (changes requested):
Thanks Luca! We're almost there! There are still a couple of places where the \` shouldn't be escaped ;-)

(Remember to squash changes in a single commit, as you've done now!) And then, it will be ready to get integrated ;-)

lucaboesch:
There you go!

sarjona (approved):

PR #97, opened 2022-05-03
[repo] Speeling from moodle

andrewnicols:
This change introduces cspell lists from two places:

- `git log --format='%aN'` to get a list of all contributors to Moodle. This is then:
  - sorted and uniqued
  - filtered to remove any name containing brackets (there are a few with `(` and `)` in them)
  - broken into new lines for each new word (that is `Andrew Lyons` becomes `Andrew\nLyons`)
  - filtered to remove any name shorter than three characters
  - sorted and uniqued once more
- a list of all Moodle plugintypes subsystems from the `components.json` file
  - Note: This does *NOT* include the list of all Moodle plugins. That requires a little more thought and possibly a Moodle installation which I am trying to avoid for as long as possible

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/97"></a>


sarjona (approved):
Hi Andrew!
That's a great improvement because will save us time to keep the list updated! Besides, it's a good idea to have separate files for components and contributors. Well done! Thanks for thinking about that and implementing it! :-)

PR #98, opened 2022-05-03
[repo] Ensure that mdx files are linted too

andrewnicols:
Markdown lint should be able to check mdx files too.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/98"></a>


sarjona (approved):
Hi Andrew!
This change makes sense so I'm happy to merge it! Thanks for raising and fixing it! :-)

PR #102, opened 2022-05-03
[docs] ' instead of ’.

lucaboesch:

PR #104, opened 2022-05-03
[repo] Add markdown lint rule to convert directional quotes to their non-directional variants

andrewnicols:
Related to #102

This should prevent additional use of the following characters:

- `“`
- `”`
- `‘`
- `’`

These will be converted to either `"` or `'` as appropriate.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/104"></a>


andrewnicols:
This commit adds a new, custom, markdownlint rule. It is _not_ applied automatically. It is _not_ possible to configure it to run from config from my understanding (deliberate design decision).

I think that this is ready but I'll check tomorrow.

andrewnicols:
Okay, this should now be ready.

I've rebased it on #106, which adds vscode workspace recommendations and configuration. This allows us to add the markdownlint custom rules into the vscode configuration. There is no way to do this in a nice, easy, way otherwise.

andrewnicols:
Hi @lucaboesch,

This change should prevent further issues with the use of true apostrophes in markdown files and will automatically fix them.

lucaboesch:
That is great news, @andrewnicols!

But it should probably look out for ‘ as well (like in https://github.com/moodle/devdocs/blame/main/general/development/process/testing/qa.md#L193).
Should we additionally worry about “ and ”, too (in the same line)?

Best,
Luca

andrewnicols:
Hi @lucaboesch ,

It specifically is looking for `’` - thats the point and it does pick up the issue in qa.md.

I'm thinking of extending it to fix `”` to `"` but I need to track down a probem with vscode first.

lucaboesch:
Yeah, @andrewnicols but you see, ‘ is just like ’ typographically.
You probably notice that it's basically the same, only one has the thicker end up, and the other one has the thicker end down.
I think that one should be looked for, too.
But, of course, vscode problem first.

andrewnicols:
Right, thanks Luca, I worked that out a short while ago. IT doesn't help that github eats half the characers unless you put them in backticks. Please ensure that you put these in the relevant backticks in future.

andrewnicols:
Okay, this is finally ready for review now that I've worked out all of the kinks.

I was looking at moving to markdownlint-cli2, but that's just a different level of pain.

lucaboesch:
Brilliant, @andrewnicols !

andrewnicols:
This has been waiting for review for a week and I really need the infrastructure changes that this provides for some related chagnes. Self-merging when tests finish.

sarjona (approved):

sarjona:
Hi @andrewnicols!
My apologies for the late review (I've been busy these days). The patch looks great and it's working as expected.
I've found a few more occurrences with this characters, but I'll create a separate issue for them :-)

PR #105, opened 2022-05-03
Migrate Moodle App release notes pages

crazyserver:
Those are all the current release note pages existing in the Dev wiki.

This can be integrated but we have to think about where to place them later, or decide it before integrating that.

However @andrewnicols has some ideas on where to place this pages and probably we should wait for them to be applied.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/105"></a>


sarjona (changes requested):
Hi @crazyserver!
Thanks for working on this. I was wondering why the Moodle Apps releases pages don't follow the same format as the [Moodle releases](https://moodle.github.io/devdocs/general/releases)?
For example, the Moodle releases the main page contains an index with all the versions: https://moodle.github.io/devdocs/general/releases
And then each release page contains only the specific things for it. So, the version folder (for instance, Moodle 4.0), is also a page with the information of the first version in the series: https://moodle.github.io/devdocs/general/releases/4.0

That way, although version numbers between app and LMS are slightly different, the releases pages will be more consistent.

general/app_releases.md
```diff
@@ -0,0 +1,80 @@
...
+
+Version 3 has been aligned with the Moodle LMS deployments.
+
+- From 3.0.0 to 3.4.2 it was based on Ionic 1 and supported Moodle 2.4 onwards.
+- On 3.5.0 name changed from Moodle Mobile to Moodle App. It was based on Ionic 3 and it was required to use Moodle 3.1 onwards.
+- On 3.9.5 the app was based on Ionic 5 version.
+
+| **Version name** | **Date** | **Version code** |
+|---|---|---|
+| [Moodle App 3.9.5](./v3/v3.9.5) | 27 August 2021 | 39500 |
```
andrewnicols, line 26: This needs to either be `./app_releases/v3/v3.9.5.md` _or_ this file needs to move to `./app_releases/index.md` as appropriate.

andrewnicols:
Just thinking that, int he long run (now now), we should create a versions.json file for the app like we have for Moodle core and we can then re-use the same infrastructure that we've put into the LMS release notes.

general/app_releases.md
crazyserver: Oh I'll check it ASAP Thanks for noticing. I've tried to move the file to index but somehow it didn't work, probably I was missing something. Anyway I'll review it.

andrewnicols: You have probably encountered the case where you cannot have an auto-generated directory with an index.md which you want to have as the _category_ page and not a member of the auto-generated sidebar.

Unfortunately that's a known limitation, and it's why the LMS release notes have the major version in the parent directory.

I do think that, ultimartely, the thing to do will be to have a `appVersions.json` file in the `data` directory with the below information, and to automatically generate this table, and the sidebar.

crazyserver: Hi Andrew, I've changed the links. However about creating the json I would open a new issue since I have no time to do it right now. Do you mind?

Thanks!!!

sarjona (approved):

PR #111, opened 2022-05-06
[repo] Automatically protect migrated pages

andrewnicols:
This commit updates our package config to use the latest version of [nodemw](https://github.com/macbre/nodemw) which adds support for protecting page (which I wrote).

The patch then adds protection for all migrated pages.

I have now added a (crappy looking) banner to the old pages linking to the new. Locking those pages prevents most people from changing them. We do still need to keep an eye on them (and maybe we can automate this better too) as admins can still make changes. We also have a few which have been updated since we migrated and need to check those and update the new docs too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "js-yaml": "^4.1.0",
     "mdx-mermaid": "^1.2.2",
     "mermaid": "^8.14.0",
-    "nodemw": "^0.16.0",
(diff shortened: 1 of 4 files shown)


sarjona (approved):
Thanks for working on this! This is a good improvement and will help us with the migration process!! <3 <3

PR #120, opened 2022-05-08
[docs] alt attribute needs no code markup.

lucaboesch:


andrewnicols:
Hi Luca,

Why don't you think that this _code example_ should have code markup?

lucaboesch:
Well, we could as well leave it in, but then, the "title" should be with code markup, too, shouldn't it?

Best,
Luca

andrewnicols:
I'd say that all html tags shown in text should be in code blocks as a policy. In this page that includs things like `"role"` `"presentation"` and `aria-hidden="true"`.

At the moment most of the _content_ of these documents has been reviewed and updated, but some typography has not been.

lucaboesch:
Yeah, you can close, @sarjona.
I was thinking that when just describing a HTML tag's attribute, it should not have to be printed like input code (after all, we're referencing the attribute's name, not its content).
But clearly my opinion has been overruled.

sarjona:
Thanks for confirming it, Luca! :-*

PR #121, opened 2022-05-10
Rewrite conversion

andrewnicols:
Originally it seemed like an awesome idea to leave the legacy `[[wikilinks]]` markdown in place during migration, however I've been advised against making up our own markdown syntax by the folks over at docusuarus.

As a result, I've been working with our own markdownlint custom rules to perform this migration as a one-off translation, and separate custom rules to detect:

- migrated pages; and
- obsolete pages.

These new rules are just checking regular links which point to `https://docs.moodle.org/dev/`.

Doing it this way means that we have one true source, which is correct in markdown always, and whicih works even when the original document was not using wikilink syntax.

As a result, we need to update the wikimedia fetch tool to use this, instead of rehype + remark + unified + some craply string replacements. This is a good thing!!

The original migration was flaky and based on regex. It was not syntax aware, and caused things like code sections to break if they included syntax-like content.

This new approach makes use of Custom markdownlint rules with `markdownlint-cli2` to make incremental syntax-aware changes to the markup.

This means that things like `=` and `#` and `''` and more, in code blocks do not break.

This still does not introduce a solution for converting uses of wikimedia templates, or wikimedia tables in the source content, and this may not be easily possible whichever approach we take, but it does make most conversions work much better.

To make this possible we've switched to using `markdownlint-cli2` (away from `markdownlint-cli`). There's a few factors behind this:

- `markdownlint-cli` only accepts custom rule configuration as a CLI argumenet (`markdownlint --rules .path/to/custom/rule`). This _sucks_ and is difficult to maintain
- `markdownlint-cli2` supports better configuration options
- `markdownlint-cli2` is used internally by the vscode markdownlint extension so we have fewer variables to handle
- `markdownlint-cli2` is written by the author of the `markdownlint` library

I've also had to fix a few issues in both the package.json and github workflows to push this.

Please note that this is currently built on #104 which is awaiting review separately.

This changeset also fices/introduces some new commands:

- `mdlint` - lint a specific document
- `mdlint-all` - lint all markdown
- `mdfix` - lint and fix a specific document
- `mdfix-all` - lint and fix all markdown

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/121"></a>


andrewnicols:
I've added the change to tracker links, and also the table migration discussed.

Self-merging so I can bulk move a load of docs.

PR #138, opened 2022-05-20
[docs] Reactive UI guide

ferranrecio:


docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+
+```handlebars title="path/to/plugin/templates/yourthing/childcomponent.mustache"
+<div id="{{uniqid}}-your-plugin-{{id}}"> (...) </div>
+{{#js}}
+require(['YOUR_PLUGIN/local/yourthing/childcomponent'], function(component) {
+    component.init('{{uniqid}}-your-plugin-{{id}}');
+});
+{{/js}}
+```
+
```
andrewnicols, line 201: It might be worth adding a tip admonitino to expalin the use of uniqid here.

```suggestion

:::tip Generating unique id attributes

You can use the `{{uniqid}}` Mustache helper within your code to help you generate a unique id for your HTML attributes and target them in your React component.

Please note that the `{{uniqid}}` helper generates a single value each time it is rendered, and you will need to combine it with other data to create a truly unique value. In this example the name of the plugin, and a static element `id` is used to generate a unique value.

:::

```

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+  </TabItem>
+</Tabs>
+
+The init method has an id as a first param to find the main DOM Element, but it also has a second one called `selectors`. This second param is an optional (but recommended) param any component can use. We will see how to use the selectors later but they are CSS selectors to find inner DOM elements in the main element. Bypassing as an init-param, any mustache can customize the default CSS selectors, which means the component is more reusable.
```
andrewnicols, line 205: I'd also put this into an `::: info` admonition. It is relevant, but will be discussed in more detail later on. This mention is just a "Heads-up" to let you the reader know that it will be covered in more detail later so hold that quesiton til then.

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+All components are registered in a Reactive Instance and, when this is done, the Component can start watching the state and ask for state mutations. However, during its life, a Component can operate differently if necessary.
+
+To understand how a component works you need to understand the component lifecycle and its hook methods:
+
+![Component lifecycle hooks](./_files/lifecycle_hooks.svg)
+
+<Tabs>
+  <TabItem value="create" label="create" default>
+
+Before the component is registered into the Reactive Instance. This method is used to define the main component attributes. At this moment the Component cannot interact with the Reactive Instance yet.
```
andrewnicols, line 218: The first sentence is not a complete sentence. Either turn it into one, or make it a piece of structured data, for example:

```suggestion
Called: Before the component is registered to the Reactive Instance.

This method is used to define the main component attributes. The Component cannot interact with the Reactive Instance yet as instantiation has not yet completed.
```

The above makes it clear where it is called in the lifecycle, and also adds a brief explanation as to why it cannot interact yet.

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+
+#### The `create(descriptor)` hook
+
+The BaserComponent construct method will register the component in the Reactive Instance after some validations. However, it is quite common for components to have private attributes needed for the UI logic. Components can set their own attributes in the `create` hook method.
```
andrewnicols, line 222: Typo: `BaserComponent`

andrewnicols:
```suggestion
The `BaseComponent` construct method will register the component in the Reactive Instance after some validation. It is quite common for components to have private attributes needed for the UI logic. Components can set their own attributes in the `create` hook method.
```

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+
+At the moment the Reactive Instance is registering the component and asking it what parts of the state it is watching. This method returns a list of state changes the component is watching and the methods that will handle those changes, something like an event listener for state changes.
+
+#### The `getWatchers(): array` hook
+
+This hook is called when the Reactive Instance is registering the component. At this moment the component can specify which state changes are watching and what methods will use to listen to those changes.
+
+The method should return an array of objects containing:
+watch: the state changes to watch. For example `user.username:update`
+handler: the class method to handle the state change event. For example `this._refreshUsername`.
```
andrewnicols, line 277:
```suggestion
The method should return an array of objects containing, with each object containing:

- watch: the state changes to watch. For example `user.username:update`
- handler: the class method to handle the state change event. For example `this._refreshUsername`.
```

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+        {watch: `section:deleted`, handler: this._deleteSection},
+    ];
+}
+```
+
+In case the component only watches changes in a specific element of a state collection, the function could be something like.
+
+```js
+getWatchers() {
+    const id = this.element.dataset.id;
```
andrewnicols, line 300: It may be worth mentioning where the `this.element` is specified elsewhere in the code and how it comes to be here.

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+        {watch: `section[${id}]:deleted`, handler: this._deleteSection},
+    ];
+}
+```
+
+  </div>
+</details>
+
+:::note
+
```
andrewnicols, line 314: Maybe add a first para like:

```suggestion

No need to manually `bind` or use the `call` function!

```

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+The Reactive Instance will call this hook on the registered components when the state data is initially loaded. This is one of the most important hooks in the Component lifecycle. At this moment the component will get its state and it can ensure that the Reactive Instance is ready and its watchers are working.
+
+Even if the component is registered once the initial state is loaded, the Reactive Instance will trigger that method when the Component is fully registered and the page is ready. For all purposes, this method is the equivalent of a page-ready event for components.
+
+The typical actions for a stateReady are:
+Add event listeners to capture user interactions.
+Refresh parts of the main element using the state data.
+Delegate logic to subcomponents.
+
+Go to the section [BaseComponent helpers](#basecomponent-helpers) to learn about some useful methods for simplifying the stateReady code.
```
andrewnicols, line 337: We tend to use "See the " rather than "Go to the", however since this is a _future_ topic it may be btter to say "The `BaseComponent` helpers section will discuss this in more detail." or similar.

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+
+  </TabItem>
+  <TabItem value="destroy" label="destroy">
+
+Before the component is unregistered. After this method is called the Component will no longer watch the state or capture the user interactions, even if the component's main DOM element is still present on the page.
```
andrewnicols, line 342: Again, suggest key/value pairing of "Called: Before the ..."

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
+---
+title: Creating a reactive UI
```
andrewnicols, line 2: Would be good to add some tags here.

docs/guides/javascript/reactive/index.md
```diff
@@ -0,0 +1,1382 @@
...
+    const state = stateManager.state;
+    fields.extras = state.activity.extras && fields.extras;
+    return fields;
+}
+```
+
+  </div>
+</details>
+
+## Drag&drop helper component
```
andrewnicols, line 1200: Suggest spaces here:

```suggestion
## Drag & drop helper component
```

andrewnicols (changes requested):
A few minor things here - nothing too big. Good work.

docs/guides/javascript/reactive/index.md
ferranrecio: I like it. Fixed.

docs/guides/javascript/reactive/index.md
ferranrecio: Done.

docs/guides/javascript/reactive/index.md
ferranrecio: done.

docs/guides/javascript/reactive/index.md
ferranrecio: Done.

docs/guides/javascript/reactive/index.md
ferranrecio: Done.

andrewnicols (approved):

PR #139, opened 2022-05-20
[docs] Migrate filter plugins

### LINKED ISSUE #29: [docs] Migrate Filters to /docs/apis/plugintype-filter
** Which page on https://docs.moodle.org/dev needs to be migrated?**

https://docs.moodle.org/dev/Filters

** Does this page need documentation specific to a Moodle version? **

Yes

** Where do you think it should be placed? **
/docs/api/plugintype-filter

ferranrecio:
Closes #29


docs/apis/plugintypes/filter/index.md
```diff
@@ -0,0 +1,201 @@
...
+    Lang,
+    Lib,
+    VersionPHP,
+} from '../../_files';
+
+import {
+    ComponentFileSummary,
+} from '../../../_utils';
+
+Filters are a way to automatically transform content before it is output. For example:
```
andrewnicols, line 28:
```suggestion
Filters are a way to automatically transform content before it is output. Filters may be used to:
```

docs/apis/plugintypes/filter/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+- Render embedded equations to images (the TeX filter)
+- Links to media files can be automatically converted to an embedded players.
+- Mentions of glossary terms can be automatically converted to links.
+
+Filters are one of the easiest types of plugin to create.
+
```
andrewnicols, line 35:
```suggestion

Filters are applied to content passing in to the `format_string()` and `format_text()` functions, which are part of the [Output API](https://docs.moodle.org/dev/Output_functions).
```

docs/apis/plugintypes/filter/index.md
```diff
@@ -0,0 +1,201 @@
...
+## Test a filter
+
+To enable a filter go to the [filters administration screen](./) and set the filter active to "On".
+
+Filters are applied to all text that is printed with the output functions `format_text()` and `format_string()`. To see a filter in action, add some content to a label resource. When you look at that course in the course listing, you should see that your filter has transformed the text accordingly.
+
+## Filter performance
+
+Is important to note that all active filters will be called to transform every bit of text output using `format_text()`, and possibly also `format_string()`. That means that a filter plugin could cause big performance problems. It is extremely important to uses caches when a filter have to get data out of the database,
```
andrewnicols, line 99:
```suggestion
It is important to note that all active filters will be called to transform every bit of text output using `format_text()` (headers and content), and `format_string()` (headers only). As a result a filter plugin can cause big performance problems. It is extremely important to uses caches if your filter must retrieve data from the database, or other similar sources.
```

docs/apis/plugintypes/filter/index.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+  </div>
+</details>
+
+## Local configuration
+
+In addition filters can also have different configuration in each context.
+
+For example, the glossary filter could be changed so that in Forum A, you can choose to only link words from a particular glossary, say Glossary A, while in Forum B you choose to link words from Glossary B.
```
andrewnicols, line 135:
```suggestion
Filters can use different configuration depending on the context in which they are called. For example, the glossary filter can be configured such that when displayed in Forum A it only links words from a particular glossary, while in Forum B it links words from a different glossary..
```

docs/apis/plugintypes/filter/index.md
```diff
@@ -0,0 +1,201 @@
...
+
+To do that sort of thing, the filter plugin must provide a **filterlocalsettings.php** file. In it, you must define a Moodle form that is a subclass of filter_local_settings_form. In addition to the standard formslib methods, you also need to define a save_changes method.
```
andrewnicols, line 137:
```suggestion
To support this behaviour, a filter plugin must provide a `filterlocalsettings.php` file which defines a Moodle form which subclasses the `filter_local_settings_form` class. In addition to the standard formslib methods, you must also define a `save_changes` method.
```

andrewnicols (changes requested):
Only a few minor changes required

docs/apis/plugintypes/filter/index.md
ferranrecio: Thanks, a commit the suggestion.

docs/apis/plugintypes/filter/index.md
ferranrecio: Thanks, a commit the suggestion.

ferranrecio:
All changes committed. Thanks!

andrewnicols (approved):

PR #140, opened 2022-05-22
[repo] Ensure that projects.json is appropriately linted

andrewnicols:
This requires a few bigger changes:

- the existing schema was against an old spec version, but included some
  features from newer spec versions. This commit includes a bump to a
  newer standard
- conversion of the lintstagedrc from json to mjs. The lint-staged UI
  doesn't like receiving extra args so we need to remove them. This is
  the easiest way

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/140"></a>


sarjona (approved):

sarjona:
Hi Andrew! Thanks for fixing this too!! :-)

PR #144, opened 2022-05-23
Update meetings.md

wildgirl:
Add link to June meeting

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/144"></a>

### DIFF
--- a/general/community/meetings.md
+++ b/general/community/meetings.md
@@ -9,13 +9,14 @@ tags:
 
 Developer meetings are open to anyone interested in Moodle development.
 
-Next meeting: [Developer meeting April 2022](https://docs.moodle.org/dev/Developer_meeting_April_2022)
+Next meeting: [June 2022](./meetings/202206)
 
 ## Past meeting notes
 
 ### 2022
 
-- [February 2022](./meetings/202202)
+- [Developer meeting April 2022](https://docs.moodle.org/dev/Developer_meeting_April_2022)
+- [Developer meeting February 2022](https://docs.moodle.org/dev/Developer_meeting_February_2022)
 
 ### 2021
 


general/community/meetings.md
```diff
@@ -9,13 +9,14 @@ tags:
...
-Next meeting: [Developer meeting April 2022](https://docs.moodle.org/dev/Developer_meeting_April_2022)
+Next meeting: [June 2022](./meetings/202206)
 
 ## Past meeting notes
 
 ### 2022
 
-- [February 2022](./meetings/202202)
+- [Developer meeting April 2022](https://docs.moodle.org/dev/Developer_meeting_April_2022)
+- [Developer meeting February 2022](https://docs.moodle.org/dev/Developer_meeting_February_2022)
```
sarjona, line 19:
```suggestion
- [Developer meeting February 2022](/general/community/meetings/202202)
```

sarjona (commented):
Hi @wildgirl !
Thanks a lot for creating this page for June's meeting :-) 
As you'll see on the main page, one of the checks is failing because the file you created is pointing to a page that has been migrated. 
I've added a suggestion with the expected code.

Apart from that, it would be nice to also migrate the page for the April's meeting (that way we'll have all the 2022 meeting pages migrated) ;-)

sarjona (changes requested):
I forgot to change the status to "Request changes". Required changes are explained in the previous comment :-)

general/community/meetings.md
```diff
@@ -9,13 +9,14 @@ tags:
 
 Developer meetings are open to anyone interested in Moodle development.
 
-Next meeting: [Developer meeting April 2022](https://docs.moodle.org/dev/Developer_meeting_April_2022)
+Next meeting: [June 2022](./meetings/202206)
```
sarjona, line 12: You need to create the general/community/meetings/202206.md too and add the content for the June's meeting :-) You can duplicate general/community/meetings/202202.md and then adapt the content (I guess that will be the easier way for doing it because in that case, there is no page to migrate).

If a page has to be migrated (for instance, April's one), you can run the following command in your devdocs cloned project and then check that the content is OK:

```yarn migrate Developer_meeting_April_2022 /general/community/meetings/202204.md```

If you haven't downloaded/installed this project locally, let me know and I'll create these files for you :-)

sarjona:
> Q: Are the ".md" suffixes needed or does it work without them? Just learning, heh.

I would say they are required (at least they appear in all the docusarus docs): https://docusaurus.io/docs/docs-introduction#document-id

andrewnicols:
> Q: Are the ".md" suffixes needed or does it work without them? Just learning, heh.

See the docs here: https://docusaurus.io/docs/next/markdown-features/links

Essentially, there are three options:

- absolute URL links (relative to the root of the project). These use the `slug` of the document. Examples include:
  - /docs/apis/plugintypes/filter
  - /general/development/policies/codingstyle
  - /
  - /versions
- a relative URL. That is a path, which does _NOT_ have the `.md` suffix, and which points tot he `slug` of the document. Examples include:
  - `../plugintypes/filter` (from `/docs/apis/subsystems/output` for example)
  - `../policies/codingstyle` (from `/general/development/proces/peer-review` for example)
- a relative FILE path. That is a path which *does* have the `.md` suffix. This points to the relative _file_ for another page. Examples include:
  - `../plugintypes/filter.md` (from `/docs/apis/subsystems/output` for example)
  - `../policies/codingstyle/index.md` (from `/general/development/proces/peer-review` for example)

Generally speaking we want to avoid relative URLs, but use either relative PATHs or absolute URLs and there are a few things to consider - in particular that we use two copies of the `@docusaurus/plugin-content-docs` plugin in Docusaurus. That's the plugin which helps us render our docs. These are the `docs` instance, and the `general` instance. We use `docs` for API guides and other version-specific Moodle platform documentation. The `general` instance is for content which is cross-version, including things like our Process docs, Policies, Coding style, Developer meeting agendas and minutes, Community information, Project documentation (because projects often last mulitple Moodle versions and the specs are useful in an ongoing capacity), and some other similar thigns.

In those versioned docs we *should* use relative file paths wherever we refer to documentation within the same Moodle version - i.e. if you are linking `/docs/apis/foo` to `/docs/apis/bar` you should use `./bar.md`. This ensures that when we generate multiple versions of the documentation that the link points to the _same_ version. If we used an absolute link (`/docs/apis/bar`) then it would always point to the version for our `master` branch. Sometimes this is desirable, but usually not.

In the unversioned docs then it generally does not matter, however, there are some other benefits to using the relative file paths, including:

- Links will keep working on the GitHub interface and many Markdown editors
- You can customize the files' slugs without having to update all the links
- Moving files around the folders can be tracked by your editor, and some editors may automatically update file links

It is not possible to use a _relative_ link to link from the `general` docs to the versioned `docs` docs.

PR #153, opened 2022-05-26
Netlify support

andrewnicols:
This commit adds (hopefully) enough to support fast and efficient builds with Netlify on a consistent build architecture.

This will be useful for us to be able to quickly preview pull request builds without having to build locally, or in GitPod, and in such a way that can be shared with others.

This also fixes an errant console.log in the Header compiler.


sarjona (approved):
Hi Andrew!
Thanks for adding support to nice tools such as Netlify (and also for removing the console logs!). The patch looks good and seems to be working properly ;-)

PR #163, opened 2022-05-30
[docs] general/process: Add apidoc build changes

andrewnicols:
One week before a major release we should update the list of branches in both of the apidoc repositories to only include the versions that will be supported after the release.

Jobs only run on a Sunday evening, so as long as the schedule is followed, the new documentation will only be generated after the branch is created for the new version, and the outgoing version will be documented until a few hours before the release.

This branch also includes a fix for cell alignment on this page.


general/development/process/release.md
```diff
@@ -100,6 +100,7 @@ tags:
...
 | 13. | &#10003; |  | Go through all points listed under 1 week prior in [New docs version process](/general/development/process/release/newuserdocs). | Community Manager |
+| 14. | &#10003; |  | Update the `BRANCHLIST` and `VERSIONLIST` in the [JSDoc](https://github.com/moodle/jsdoc) and [PHPDoc](https://github.com/moodle/phpdoc) repositories with the list of supported versions. These are built on a Sunday so this should be done in the final week before release. | Integration Team |
```
sarjona, line 103:
```suggestion
| 14. | &#10003; | &#10003; | Update the `BRANCHLIST` and `VERSIONLIST` in the [JSDoc](https://github.com/moodle/jsdoc) and [PHPDoc](https://github.com/moodle/phpdoc) repositories with the list of supported versions. These are built on a Sunday so this should be done in the final week before release. | Integration Team |
```

sarjona: As we also need to remove unsupported versions from the list, shouldn't this action be checked also before minors (as we do with some similar maintenance tasks related to unsupported branches?)

sarjona (changes requested):
Hi @andrewnicols!
The change looks great but, as you'll see, I have a minor change.

BTW, I was searching for a way to align content in tables but I wasn't able to find it! So I'm happy because with your patch I've also learnt a new thing! Thanks!! :-*

general/development/process/release.md
andrewnicols: Yup - you're right.

sarjona (approved):
Great, thanks! :-)

PR #164, opened 2022-05-31
[docs] Update front page with feedback from marketing team

andrewnicols:
This is a WIP, but so far uncorporates:

- Adopt .com-style navbar
- Switch logo (need one with TM apparently but have not got this yet)
- switch to cards (awaiting images or mockups for desired look)
- switch title to new style with em
- ditch orange hero banner
- switch font to Open Sans for most content
- switch front to DM Sans Serif for hero banner titles

Hopefully we'll have some copy, images and final mockups soon.

If we can get the copy sooner then I'll land this change as-is and we can sort out the other pieces later.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/164"></a>


src/components/HomepageFeatures/index.js
```diff
@@ -42,34 +42,54 @@ const FeatureList = [
...
+        title: 'Release notes',
         Svg: require('@site/static/img/undraw_docusaurus_react.svg').default,
         description: (
             <>
-                Discover our community, and learn about our mission, and our roadmap.
+                Technical specs and new features of Moodle releases, supported versions and more.
             </>
         ),
-        linkText: 'Discover Communities',
+        linkText: 'Check the Release notes',
```
sarjona, line 81:
```suggestion
        linkText: 'Check the release notes',
```

src/components/HomepageFeatures/index.js
```diff
@@ -42,34 +42,54 @@ const FeatureList = [
         Svg: require('@site/static/img/undraw_docusaurus_tree.svg').default,
         description: (
             <>
-                Learn about our coding standards, styles, and conventions.
+                Read our style guidelines, including coding standards and naming conventions.
             </>
         ),
-        linkText: 'Learn more',
+        linkText: 'View Standards',
```
sarjona, line 48:
```suggestion
        linkText: 'View standards',
```

sarjona (changes requested):
The patch looks good. Thanks for working on it and making it seems like moodle.com. Well done! :-)

As you'll see, there are only a minor uppercases that I think should be moved to lowercase.

Apart from that, I hope we receive the design soon. Although the current one looks more similar to moodle.com, I'm still not convinced with the colours and I think we need a different logo, a mark or anything to differentiate moodle.com from Moodle devdocs.

src/components/HomepageFeatures/index.js
andrewnicols: I disagree with this one. "Release notes" are a proper noun, they're not just another word. They should therefore be capitalised.

andrewnicols: Done.

PR #166, opened 2022-05-31
[docs] Migrade Upgrade notes page

sarjona:
Migrated page:
- https://docs.moodle.org/dev/Upgrade_notes

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/166"></a>

### DIFF
--- a/docs/apis/_files/upgrade-txt.mdx
+++ b/docs/apis/_files/upgrade-txt.mdx
@@ -1,0 +1,17 @@
+<!-- markdownlint-disable first-line-heading -->
+Each component and subsystem may make use of an `upgrade.txt` file in the top level folder. A title is used to identify the Moodle version where the change was introduced, and significant changes for that version relating to that component or subsystem are displayed.
+
+When changes are integrated to multiple branches, for example a stable version and the master branch, then the relevant versions used to describe the change in the `upgrade.txt` file should be the next version to be released _for each branch_. The `master` branch should always use the next major version.
+
+For example if a change is applied to the **MOODLE_400_STABLE** during the development of Moodle 4.0.2, and the **master** branch during the development of Moodle 4.1, then the relevant versions will be **4.0.2** and **4.1**.
+
+If a change is only applied to the master branch then only the **4.1** version will be listed.
+
+:::note Exception during parallel development
+
+When Moodle is developing two major versions in parallel, for example Moodle 3.11.0, and Moodle 4.0.0, then the
+version in the earliest of the major version development branches will be used for both branches.
+
+For example, during the development of Moodle 3.11, and Moodle 4.0, which took place in parallel, if a change was integrated to 3.11 and 4.0, then the version specified in both branches would be 3.11.
+
+:::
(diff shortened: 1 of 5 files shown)


docs/apis/_files/upgrade-txt.mdx
```diff
@@ -0,0 +1,15 @@
...
+When changes are integrated to multiple development branches (for example, two major releases), and they are commented in the `upgrade.txt` notes, **the version listed will be the earliest development branch that the change was made to**. Besides, the master `upgrade.txt` files should only have major versions.
+
+For example:
+
+- In **normal development period**, such as during the 4.1 development, if a change is applied to 4.0.2 and master (4.1), then the `upgrade.txt` notes will show **4.0.2** in *MOODLE_400_STABLE* and **4.1** in *master*.
+- During the 3.11 development, which was **developed in parallel** with the master branch for Moodle 4.0:
+  - if a change is applied to 3.11 and master then the `upgrade.txt` notes will show **3.11** in both branches
+  - if a change was made to 3.9.3, 3.11, and master (4.0) then that change will be marked as **3.9.3** on the *MOODLE_39_STABLE* branch, and **3.11** for both *MOODLE_311_STABLE* and *master*.
+
+:::
```
andrewnicols, line 15: I find this all very confusing and I don't understand it ;)

```suggestion

When changes are integrated to multiple branches, for example a stable version and the master branch, then the version used to describe it in the `upgrade.txt` file should be the next version to be released _for that branch_. The `master` branch should always use the major version.

For example if a change is applied to the **MOODLE_400_STABLE** during the development of Moodle 4.0.2, and the **master** branch during the development of Moodle 4.1, then the `upgrade.txt` notes will show **4.0.2** in *MOODLE_400_STABLE* and **4.1** in *master*.

If a change is only applied to the master branch then only the **4.1** version will be listed.

:::note Exception during parallel development

When Moodle is developing two major versions in parallel, for example Moodle 3.11.0, and Moodle 4.0.0, then the 
version in the earliest of the major version development branches will be used for both branches.

For example, during the development of Moodle 3.11, and Moodle 4.0, which took place in parallel, if a change was integrated to 3.11 and 4.0, then the version specified in both branches would be 3.11.

:::
```

andrewnicols (changes requested):

docs/apis/_files/upgrade-txt.mdx
andrewnicols:
```
When changes are integrated to multiple branches, for example a stable version and the master branch, then the relevant versions used to describe the change in the `upgrade.txt` file should be the next version to be released _for each branch_. The `master` branch should always use the next major version.

For example if a change is applied to the **MOODLE_400_STABLE** during the development of Moodle 4.0.2, and the **master** branch during the development of Moodle 4.1, then the relevant versions will be **4.0.2** and **4.1**.
```

sarjona:
I've updated the text with the latest proposal! Thanks Andrew for helping us to improve this part too! <3

andrewnicols (approved):

PR #182, opened 2022-06-07
[repo] Switch to commit-specific wait-for-netlify alternative

andrewnicols:
The previous copy of wait-for-netlify is only looking for any build, not for a specific build. That meant that subsequent pushes to the branch on the same pull request immediately return the previous Netlify build, and the lighthouse report related to that build.

This change also updates the Lighthouse report to run against the permalink for the build rather than the Deploy preview. The difference is that the permalink is the pure build as it would be against on the live site, whilst the deploy preview includes a bottom banner for Netlify with some useful tools. This negatively affects the performance scores for Lighthouse and brings up some false issues.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/182"></a>

### DIFF
--- a/.github/workflows/lighthouse-report.yml
+++ b/.github/workflows/lighthouse-report.yml
@@ -16,26 +16,30 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3
+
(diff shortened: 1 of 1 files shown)


sarjona (approved):
Thanks for working on this :-)

PR #183, opened 2022-06-07
[repo] Correct old Github IO URIs to moodledev.io

andrewnicols:


sarjona (approved):
Thanks for taking care of this and updating the old references. The patch looks good

BTW, I've also edited the "Migrated" template in Mediawiki (https://docs.moodle.org/dev/Template:Migrated) to update this URL too.

PR #193, opened 2022-06-07
Drag and drop

lucaboesch:


andrewnicols:
Thanks Luca,

I amended the commit message to prefix it with `[docs]: Fix typo`  to make it clearer what the change was about.
Manually merged on the CLI so this issue was not automatically closed.

PR #194, opened 2022-06-08
[docs] Spell JavaScript correctly.

### LINKED ISSUE #191: [docs] No standard whether to write javascript, Javascript or JavaScript

https://moodledev.io/general/development/policies#javascript

You see the expression javascript, Javascript or JavaScript written differently through the whole of the docs.
Of course this can't be changed in the titles of historic issues, but it would be nice if it was written consistently in the manual texts.

The paragraph literally is titled "JavaScript" with the first line being "New Javascript in Moodle should be written as Vanilla javascript in the ES6 style." 🤣
Three different ways of spelling in two lines.

No (Default)

lucaboesch:
Fixes #191

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/194"></a>


andrewnicols (approved):

andrewnicols:
Thanks Luca,

We would be better to handle this by writing a markdownlint fix to prevent future recurrences. I'm just testing a solution which does so.

PR #195, opened 2022-06-08
Canonicalise word capitalisation

andrewnicols:
@lucaboesch this is the change I was proposing.

This patch introduces a new markdownint rule to look for incorrect casing of specified words and correc them.

It does not apply in fenced code blocks, nor inline code blocks, or the link component of a link. It does apply to the description of a link.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/195"></a>


sarjona (approved):
Thanks for adding this rule to make docs more consistent :-) 
The patch looks good!

PR #198, opened 2022-06-08
[repo] Tidy up Workflow names to make them clearer

andrewnicols:


.github/workflows/lighthouse-report.yml
```diff
@@ -1,4 +1,4 @@
-name: Lighthouse Report
+name: "CI - Lighthouse Report"
```
timhunt, line 1: This change does nothing to solve the original problem, which is that if you have never heard of 'Lighthouse' then this name is meaningless.

How about 

"CI - Lighthouse website audit"

It is a pity that workflows can't have a Description: to explain them, unlike some other things in the .yaml file.

timhunt (commented):
These are all good improvements, but I think you have failed to solve the originally reported problem. See specific comment.

.github/workflows/lighthouse-report.yml
andrewnicols: I know right! I wish we could have descriptions here

andrewnicols:
I've renamed this workflow per your suggestion @timhunt.

timhunt (approved):
Looks good now. I think this can be merged.

andrewnicols:
Thanks Tim

PR #204, opened 2022-06-10
[docs] Migrate Groups docs

andrewnicols:
Migrate the groups subsystem docs.

Note: There was previously a migrated doc for groups, but it turned out to be a bad and mostly empty placeholder for a long-abandoned project with out-of-date information.

I've adjusted this so that it will now point to the correct docs.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/204"></a>


data/migratedPages.yml
```diff
@@ -122,8 +122,11 @@ Filters:
 - filePath: "/docs/apis/plugintypes/filter/index.md"
   slug: "/docs/apis/plugintypes/filter/"
 Groups:
-- filePath: "/docs/guides/groups.md"
-  slug: "/docs/apis/subsystems/groups"
+- filePath: "/docs/guides/group/index.md"
+  slug: "/docs/apis/subsystems/group"
```
ferranrecio, line 126: Why "guides/groups.md" is defined here if it does not exists?

docs/apis/subsystems/group/index.md
```diff
@@ -0,0 +1,140 @@
...
+$discussiongroup = $DB->get_record('forum_discussions', ['id' => $discussionid], groupid);
+
+// Check access.
+if (groups_get_activity_groupmode($cm)) {
+    $groups = groups_get_activity_allowed_groups($cm);
+} else {
+    $groups = [];
+}
+if (!in_array($discussiongroup->groupid, array_keys($groups))) {
+    print_error('groupnotamember', 'group');
```
ferranrecio, line 107: I found this example a bit confusing. If the forum is in group mode "NOGROUPS" (value zero) it won't let the user access it.

andrewnicols: If it's in no group mode, then the user shoudl not have access to discussions which have a group b.

data/migratedPages.yml
andrewnicols: Ah - that's a mistake. I moved it after original creation and missed the filePath update.

docs/apis/subsystems/group/index.md
ferranrecio: I think it is a strange example. If it checks the group mode only if the post has a group id it will feel more real.

ferranrecio:
I create two extra commits with some small corrections. My idea was to do it as a comment but I click the wrong button and now I am not able to remove those commits (sorry, mi first PR using GitHub UI). The corrections there are part of the PR but feel free to squash or drop them.

docs/apis/subsystems/group/index.md
andrewnicols: But that would be wrong. If a discussion has a group, and I am not in that group, then I should not be able to access it (except with the accessallgroups capability).

In any case, this is a long standing and existing example. It is 100% correct and was in the codebase until about 12 months ago when the code structure was changed but the same logic is still in place.

Remember that a discussion may have been in a group and now the course reconfigured. It would be wrong to display that discussion to users to whom it was not originally intended

andrewnicols:
I've squashed those commits, rebased, and force pushed.

docs/apis/subsystems/group/index.md
ferranrecio: OK then. I miss some examples using the group mode constants but this is something we can add later. There is no reason to delay this pull request for this.

ferranrecio (approved):

PR #207, opened 2022-06-10
[repo] Perform mdfix-all after migrating pages

andrewnicols:


sarjona (approved):
Thanks for adding this improvement to the migration tool :-)

PR #210, opened 2022-06-10
[repo] Switch to Algolia search

andrewnicols:
We have an Open Source license for Algolia DocSearch which gives us free
use of Algolia.


andrewnicols:
I've worked with Algolia to update our domain name to fix the Algolia configuration. This is now working.

sarjona (approved):
I love the new Algolia search engine! It works locally and displays the information nicely!! Thanks for taking care of this!! <3

PR #203, opened 2022-06-10
[docs] Migrate DML Documentation

andrewnicols:
This commit only migrates the DML, DDL, and SQL Coding style.

The XMLDB documentation is probably best treated as a separate migration
as it is quite significant in size.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/203"></a>

PR #211, opened 2022-06-10
[docs] Migrate Form API documentation

andrewnicols:
This is a good start to the migration of the Forms API documentation.

There's still a fair bit to do, but I think it best to do this as a separate commit to preserve sanity. Most of the remaining documentation is either long and complex (may need splitting up), or so woefully out-of-date that it will require a lot of time to update.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/211"></a>


andrewnicols:
Note: I have deliberately lef the `core` folder empty as it will use the folder created by #203.

andrewnicols:
Note: I originally had this in /docs/apis/core but moved it to /docs/apis/subsystems

Rather than hide history for this I've written a rename tool to update all links. I felt it better to do this now, and include it here as I used it as part of the rename. This demonstrates. how it works, and gives examples of its usage.

andrewnicols:
I'm going to self-merge this as it's been waiting for over a week now.

PR #213, opened 2022-06-11
[docs] docs: Tidy up the Guides landing

andrewnicols:
This changes adds a new guides landing page for the non-API guides, and
updates the main landing page for the version-specific documentation.

I've also removed the link to the 'Getting started' from the sidebar
because it takes you away from the API information and changes the
sidebar navigation completely, which makes it harder to go back to the
previous area.

As part of this I've also restructured the sidebar to move the Developer
update notes to the top section above the break, so that all
documentation for the version is together, in one place.

### DIFF
--- a/config/navbar.js
+++ b/config/navbar.js
@@ -32,10 +32,9 @@ const navbar = {
     },
     items: [
         {
-            type: 'doc',
-            docId: 'apis',
-            position: 'left',
+            to: '/docs',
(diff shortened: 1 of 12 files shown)


sarjona (approved):
I like the new organisation. It's clear. Besides, it's great you've also added some extra information to the Introduction page and make some of the plugin/API pages more consistent! Well done!

PR #220, opened 2022-06-14
[docs] Migrate plugin types page index

andrewnicols:


sarjona (approved):
Thanks for migrating the plugin types page index :-) It looks good!

PR #229, opened 2022-06-15
get_record example doesn't use an array for params

sharpchi:


andrewnicols (approved):

andrewnicols:
Thanks Mark - good catch.

PR #236, opened 2022-06-20
[docs] fix reactive example

ferranrecio:


docs/guides/javascript/reactive/index.md
```diff
@@ -146,25 +146,35 @@ export default class extends BaseComponent {
...
 
-<details>
-  <summary>View example</summary>
-  <div>
+<Tabs>
+  <TabItem value="init-js-module" label="AMD module" default>
 
 ```js title="path/to/plugin/amd/src/main.js"
 import {reactiveInstance} from 'YOUR_PLUGIN/reactive';
 import {YourComponent} from 'YOUR_PLUGIN/yourcomponent;
```
PhMemmel, line 154:
```suggestion
import YourComponent from 'YOUR_PLUGIN/yourcomponent;
```
As far as I understand, here has to be imported the component itself as the component does not export a named object, but a default class? I had to change it like that in my test project to make it work.

andrewnicols (approved):

PR #241, opened 2022-06-22
Migrate undocumented plugintypes

andrewnicols:
These are just placeholders. These docs don't exist right now.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/241"></a>


andrewnicols:
Merging without review. This is non-controversial and we really need to migrate as much content over and fork for 4.1 ASAP.

PR #243, opened 2022-06-22
[repo] Noting a page as migrated is a major edit

andrewnicols:
When we started marking pages as protected we should have also set the template addition as a major change and not a minor change.


stronk7 (approved):
Looks perfect, and I think it's important information for watchers to know the page has been moved.

PR #244, opened 2022-06-22
Add some recently details that were missed on translation

stronk7:
These changes correspond to:

https://docs.moodle.org/dev/index.php?title=Github_actions_integration&type=revision&diff=63357&oldid=58361

that were performed after the GHA move PR was created. With this
landed the new dev docs should be 100% on pair with old ones.


stronk7:
(force-pushed to change some `<code>` uses by Markdown's backticks)

Both seem to work the same, but surely it's better to use the "native" alternative then the html one.

Ciao :-)

andrewnicols (approved):

PR #264, opened 2022-06-28
[docs] Autogenerate the tools list

andrewnicols:
As discussed in the DR channel.

Note: There are some other areas that I could also do this too, but they're almost all much more complex and I'd rather wait until we have a need to do so for the moment (or spend some time writing a tool to simplify the necessary renames).

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/264"></a>


general/development/tools/nodejs.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: NodeJS and Grunt
+sidebar_position: 1
```
ilyatregubov, line 3: Should this be 2? 

Do we always need to specify explicit order?

ilyatregubov (changes requested):

general/development/tools/nodejs.md
andrewnicols: Nope - it's intentionally 1 and therefore falls back to lexical sorting.

I've ordered these such that they're:
- during development (phpcs, and node)
- after development (mdk)
- automated actions on push (gha)
- automated actions during review (cibot)

Essentially trying to follow the logical order of when you'd expect to use or see them.

ilyatregubov (approved):
Works as expected. Thanks!

PR #265, opened 2022-06-28
[docs] Migrate Calendar API docs

laurentdavid:
Migrated pages:

* https://docs.moodle.org/dev/Calendar_API
* https://docs.moodle.org/dev/Calendar_API_old

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/265"></a>


docs/apis/subsystems/calendar/index.md
```diff
@@ -0,0 +1,557 @@
...
+
+```php
+[[1510625037, "This is the maximum cutoff"](null,)] // No minimum cutoff.
+```
+
+```php
+[[1510625037, “This is the minimum cutoff”], null] // No maximum cutoff.
+```
+
+If the calendar event has no valid timestart values then the callback should return [false](false,). This is used
```
andrewnicols, line 476: This is teh cause of the build failure

laurentdavid:
I have closed the PR as it was migrated since then.

PR #279, opened 2022-07-07
[docs] Migrate the "Moodle and PHP" page and related changes

stronk7:
This migrates the https://docs.moodle.org/dev/Moodle_and_PHP
page to devdocs/general/development/php and, given php80
support has been officially announced for 3.11 and 4.0,
amend the release notes of those versions to tell so.

Then, register it as migrated page.

And finally, fix all the links to the old page (`yarn mdfix-all`).

Let's see if everything is in place (`yarn build` said ok), yay!

Ciao :-)

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/279"></a>

### DIFF
--- a/data/migratedPages.yml
+++ b/data/migratedPages.yml
@@ -1197,6 +1197,9 @@ Moodle_4.0_developer_update:
 Moodle_4.0_release_notes:
 - filePath: "/general/releases/4.0.md"
   slug: "/general/releases/4.0"
+Moodle_and_PHP:
+- filePath: "/general/development/policies/php.md"
+  slug: "/general/development/php"
 Moodle_App:
 - filePath: "/docs/moodleapp.md"
   slug: "/docs/moodleapp"
--- a/general/development/policies/php.md
+++ b/general/development/policies/php.md
@@ -1,0 +1,94 @@
+---
+title: PHP
+sidebar_label: PHP versions
+description: Information about how PHP supported versions correlate with Moodle releases and the policy controlling it.
+tags:
+  - Requirements
+  - Support
+  - Releases
+  - PHP
+---
+
+import { Since } from '@site/src/components';
+
+New PHP versions are [released every year](https://www.php.net/supported-versions.php) and come with important improvements and changes compared with previous versions. Moodle tries to support them as soon as possible, always matching them with our own [scheduled release plans](/general/releases).
+
+## Policy statement
+
+<Since versions={["3.5"]} issueNumber="MDL-59159" />
+
+We always follow this agreed policy regarding PHP and Moodle supported versions:
+
+1. A LTS will always **require the previous LTS** (or later) for upgrading.
+2. The **maximum PHP version** supported for a branch will be the max one achieved along the life of the branch. Usually with .0 releases but may happen later (we added support for php74 with 3.8.3, or support for php80 with 3.11.8, for example).
+3. The **minimum PHP** version supported for a branch will be **the lower of**:
+    - The [minimum version supported in any way by php](https://www.php.net/supported-versions.php) the day of the Moodle release (so we provide slow, progressive increments).
+    - The maximum PHP version supported by the previous LTS branch (so we guarantee jumping between LTS is possible without upgrading PHP at the same time).
+
+<details><summary>PHP and Moodle policy in Jira wiki markup format.</summary>
+
+```txt
+{panel:title=Policy: PHP & Moodle supported versions|borderStyle=dashed|borderColor=#cccccc|titleBGColor=#f7d6c1|bgColor=#ffffce}
+Since Moodle 3.5 (MDL-59159), these rules apply to decide Minimum PHP and Moodle versions supported:
+ # A LTS will always require the previous LTS (or later) for upgrading.
+ # The maximum PHP version supported for a branch will be the max one achieved along the life of the branch. Usually with .0 releases but may happen later (we added support for php70 with 3.0.1, or support for php73 with 3.6.4, for example).
+ # The minimum PHP version supported for a branch will be *the lower of*:
+ -- The [minimum version supported in any way by php|http://php.net/supported-versions.php] the day of the Moodle release (so we provide slow, progressive increments).
+ -- The maximum PHP version supported by the previous LTS branch (so we guarantee jumping between LTS is possible without upgrading PHP at the same time).{panel}
+```
+
+</details>
+
+This page contains the current status of support for every PHP version and Moodle versions. For details, follow also the epic links below that combine all the changes that were made in Moodle to ensure PHP compatibility.
+
+:::note
+
+You must be logged in to tracker to see issues in Epics.
+
+:::
+
+## PHP supported versions
+
+### PHP 8.1
+
+PHP 8.1 support **is being implemented** for Moodle 4.1 and later releases. Hence it's still **incomplete and only for development purposes**.  See [MDL-73016](https://tracker.moodle.org/browse/MDL-73016) for details.
+
+### PHP 8.0
+
+<Since versions={["3.11.8", "4.0.2"]} issueNumber="MDL-70745" />
+
+PHP 8.0 **can be used with** Moodle 3.11.8, Moodle 4.0.2 and later releases. See [MDL-70745](https://tracker.moodle.org/browse/MDL-70745) for details.
+
+### PHP 7.4
+
+<Since versions={["3.8.3", "3.9"]} issueNumber="MDL-66260" />
+
+PHP 7.4 **can be used with** Moodle 3.8.3, Moodle 3.9 and later releases. It is also the **minimum** supported version for Moodle 4.1. See [MDL-66260](https://tracker.moodle.org/browse/MDL-66260) for details.
+
+### PHP 7.3
+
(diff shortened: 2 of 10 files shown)


general/development/php.md
```diff
@@ -0,0 +1,30 @@
...
+  - Support
+  - Releases
+  - PHP
+sidebar_position: -400
+---
+New PHP versions are [released every year](https://www.php.net/supported-versions.php) and come with important improvements and changes compared with previous versions. Moodle tries to support them as soon as possible, always matching them with our own [scheduled release plans](/general/releases).
+
+This page contains the current status of support for every PHP version and Moodle versions. For details, follow also the epic links below that combine all the changes that were made in Moodle to ensure PHP compatibility. *(Note: you must be logged in to tracker to see issues in Epics)*
+
+- PHP 7.0 **can be used with** Moodle 3.0.1, Moodle 3.1 and later releases. It is also the **minimum** supported version for Moodle 3.4. See [Moodle and PHP 7.0 details](https://docs.moodle.org/dev/Moodle_and_PHP_7.0_details) and [MDL-50565](https://tracker.moodle.org/browse/MDL-50565) for details.
```
andrewnicols, line 14: It may be worth putting each PHP version in its own heading. That woudl automatically generate an id on the title which can be referenced in a url, and it woudl add it to the ToC.

It _may_ also be worth adding a Since tag to the section. This will need to be imported near the top of the page:

```
import { Since } from '@site/src/components';
```

And then you can use it as a React element:

```suggestion

## PHP 7.0

<Since
    versions={["3.0.1", "3.1"]}
    issueNumber="MDL-50565"
/>

PHP 7.0 **can be used with** Moodle 3.0.1, Moodle 3.1 and later releases. It is also the **minimum** supported version for Moodle 3.4. See [Moodle and PHP 7.0 details](https://docs.moodle.org/dev/Moodle_and_PHP_7.0_details) and [MDL-50565](https://tracker.moodle.org/browse/MDL-50565) for details.
```

The source for that element is at https://github.com/moodle/devdocs/blob/main/src/components/Since.js, though I'll be rewriting it in TypeScript sometime with fully described props.

andrewnicols (changes requested):
Hi Eloy,

Thanks for migrating this page.

There's a couple of minor suggestions for you in the code, though I'm not sure about the best place to display this document in the hierarchy.

I wonder whether we should put this into the policies folder and to then also include the policy wording at the top of the document, that is (stolen from MDL-71747):

```
{panel:title=Policy: PHP & Moodle supported versions|borderStyle=dashed|borderColor=#cccccc|titleBGColor=#f7d6c1|bgColor=#ffffce}
Since Moodle 3.5 (MDL-59159), these rules apply to decide Minimum PHP and Moodle versions supported:
 # A LTS will always require the previous LTS (or later) for upgrading.
 # The maximum PHP version supported for a branch will be the max one achieved along the life of the branch. Usually with .0 releases but may happen later (we added support for php70 with 3.0.1, or support for php73 with 3.6.4, for example).
 # The minimum PHP version supported for a branch will be *the lower of*:
 -- The [minimum version supported in any way by php|http://php.net/supported-versions.php] the day of the Moodle release (so we provide slow, progressive increments).
 -- The maximum PHP version supported by the previous LTS branch (so we guarantee jumping between LTS is possible without upgrading PHP at the same time).{panel}
```

Perhaps if we do this then the document structure might be:

PHP
- Supported versions policy
  - Wording
- Supported versions
  - 8.1
  - 8.0
  - etc.

stronk7:
Thanks for the review @andrewnicols , learning a lot!

I'll go examining your suggestions and incorporating them as new commit, surely to squash everything at the end (merge and rebase by you OR I can pre-rebase locally too, as you wish).

Will ask for a new review once I'm done, still have to digest/understand some of your wise suggestions!

Ciao :-)

general/development/php.md
```diff
@@ -0,0 +1,30 @@
...
+
+- PHP 7.4 **can be used with** Moodle 3.8.3, Moodle 3.9 and later releases. It is also the **minimum** supported version for Moodle 4.1. See [MDL-66260](https://tracker.moodle.org/browse/MDL-66260) for details.
+
+- PHP 8.0 **can be used with** Moodle 3.11.8, Moodle 4.0.2 and later releases. See [MDL-70745](https://tracker.moodle.org/browse/MDL-70745) for details.
+
+- PHP 8.1 support **is being implemented** for Moodle 4.1 and later releases. Hence it's still **incomplete and only for development purposes**.  See [MDL-73016](https://tracker.moodle.org/browse/MDL-73016) for details.
+
+## See also
+
+- [Roadmap](/general/community/roadmap) - future versions
```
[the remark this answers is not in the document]
stronk7: Oh,

I think I really wanted to link it to the /general/releases (Releases) page, but surely I copied it from the wrong place! Will change that!

general/development/php.md
stronk7: Ok, interesting! I'm going to try this locally, instead of applying your suggestion, so I see it in action. Thanks!

stronk7:
Ok, I think I've achieved everything (all the changes are in 3 fixup! commits).

1) The page is now `general/development/policies/php` (and all links and the migrated-pages file have been updated)
2) It shows the policy statement and the jira-wiki version in a collapsed doc block.
3) The versions have been re-arranged, newer to older.
4) Each one has now its own level 3 heading and the "Since" badges.
5) The `sidebar_position`has been removed.
6) The see also section has been fixed to point to the releases page (that was the initial intention).

And I think that was all.

Let me know if there is anything else to change, ciao :-)

OT: I spent hours getting "CodeBlock" to work, creating an external "php_policy.txt" file and loading it using the loaders and all those things. Just to discover, once I got it working that I could use the normal markdown code backticks and done. Was fun!

stronk7:
Let me know if you want me to rebase the fixup! commits to just leave 3 commit: 1) the page, 2) the migratedFiles and 3) the links.

Or feel free to "Squash and merge" all them down to just one commit. NP here.

Re-ciao :-)

andrewnicols:
Hi Eloy,

That change looks great. The only thing I'll suggest is that we add a bit more metadata. and there are a few reasons:

Firstly we should probably add a [`sidebar_label`](https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-docs#sidebar_label) to change the title to something like "PHP Versions" in the sidebar.

Secondly we shoudl probably add a [description](https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-docs#sidebar_label) config to the frontmatter. The way that the description is guessed is to take the first few paragraphs and pass them through some regexes. Unfortunately they don't remove JSX - so because there's an `import { Since } from '@site/src/components';` that will be in the default description. It's a pain but it's an upstream issue with Docusaurus so we don't have a huge amount of control over it.

Otherwise this looks great.

If you can squash the commits, that's great, otherwise I can squash during the merge.

stronk7:
Ok, I have:

1) Squahed the fixup commits so we are back to the original 3 (new page, migratedPages and updated links)
2) Added the requested metadata (sidebar_label && description) this way:
```
diff --git a/general/development/policies/php.md b/general/development/policies/php.md
index 1c17a3088..54e0316a3 100644
--- a/general/development/policies/php.md
+++ b/general/development/policies/php.md
@@ -1,5 +1,7 @@
 ---
 title: PHP
+sidebar_label: PHP versions
+description: Information about how PHP supported versions correlate with Moodle releases and the policy controlling it.
 tags:
   - Requirements
   - Support
```

stronk7:
Side comment: Maybe it wouldn't be crazy to have some good templates (haven't seen any, maybe they exist) with the usual elements to be added (title, sidebar stuff, descriptions, tags...), the skeleton of the page (intro, main section, see also...) and maybe, also, the more often used artifacts (Since....).

Without your guidance in this issue it would have been impossible for me to know about all them (and sure that there are many others).

Little side comment: Do we really need to manually create all the MDL/MDLSITE/CONTRIB links by hand? Wouldn't be better (and more future-proof) to make that something automatically built?

Ciao :-)

andrewnicols (approved):

andrewnicols:
> Side comment: Maybe it wouldn't be crazy to have some good templates (haven't seen any, maybe they exist) with the usual elements to be added (title, sidebar stuff, descriptions, tags...), the skeleton of the page (intro, main section, see also...) and maybe, also, the more often used artifacts (Since....).
> 
> Without your guidance in this issue it would have been impossible for me to know about all them (and sure that there are many others).

Thanks for the suggestion - we should definitely do that. If someone can review #283 then I'll start on that once it lands and will work to refine that section more. I've raised #296 to document this. If there's anything else you can think to add to this pleaes comment on that issue.

> Little side comment: Do we really need to manually create all the MDL/MDLSITE/CONTRIB links by hand? Wouldn't be better (and more future-proof) to make that something automatically built?

I've been on the fence there. It is something that we _can_ do, but I don't know whether should. We can do it as a Remark plugin - in fact we originally did something similar by converting `{tracker {PROJ-XXXXX}` into tracker links but moved away from that because I was advised that making up random markdown syntax is a bad idea. It wold probably be fine to do just a link... I have raised #295 to look into doing this.

PR #280, opened 2022-07-08
[repo] Make yarn fix to call to common fixes

andrewnicols:
This pair of commit adds `yarn fix` as a shortcut to `yarn mdfix-all; yarn mdxfix-all; yarn spell` to perform standard checks for all files.
It then also updates the `yarn migrate` command to use the new command, and makes it run _after_  we add to `migratedPages.yml` which makes it actually useful in that instance.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/280"></a>

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "wikimedia-sync": "scripts/wikimedia-sync.js",
     "component-spellings": "scripts/component-helper.js spelling",
     "test": "jest",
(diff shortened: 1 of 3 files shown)


sarjona (approved):
Hi Andrew! 
Thanks for adding this new command (`yarn fix`) and improving the migrate script :-)

PR #284, opened 2022-07-08
[repo] Lint all markdown files when migratedPages changes

andrewnicols:
Note: This will fix all markdown files but it **will not** stage them
into git, or block the commit. It doesn't work quite that easily.

We could _possibly_ block the commit without the files, but it's not
trivial because there may be legitimate reasons that people want to
split commits anyway.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/284"></a>


sarjona (approved):
Hi Andrew!
The change makes sense to me :-) Thanks for working on it!

PR #281, opened 2022-07-08
Absolute links general

andrewnicols:
This enables the forcing of relative links for the general directory.

    Whilst we do not version this directory, this enables links to be
    followed in most IDEs, and allows our markdownlint rename feature to
    work.

PR #286, opened 2022-07-08
[docs] Move the moodleapp to the /general directory

### LINKED ISSUE #86: [chore] Migrate app out of `docs` directory
### What do we need to do?

Move the app docs out of the `docs` directory into its own location.

andrewnicols:
As mentioned in #86, we should move the moodleapp documentation from /docs to /general.

The /docs folder is used for documentation relating to a specific version of _Moodle_ whilst the App covers the _current time_ and multiple Moodle versions. It is a living document rather and therefore belongs in /general.

Note: This issue is based on the patch in #281 which helps with the renaming of files in this manner.

Closes #86.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/286"></a>


sarjona (approved):
Thanks Andrew for moving the app pages from /docs to /general. I'm still unsure if they should have or not a separate folder but I agree that it's better to have them in the /general folder (because they are not versioned) :-)

PR #292, opened 2022-07-08
[docs] Fix links to moodledev.io/devdocs

andrewnicols:


stronk7 (approved):
Have reviewed that all the new links do, indeed, exist, so I imagine this is correct, yay!

PR #298, opened 2022-07-11
Update release dates on the release notes

junpataleta:
In the future, perhaps it would be better to just note the planned release date when drafting the new release notes for the next minors. People should be able to determine the actual release status of the version from the indicated release date, right?

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/298"></a>

### DIFF
--- a/general/releases/3.11/3.11.8.md
+++ b/general/releases/3.11/3.11.8.md
@@ -9,7 +9,7 @@ moodleVersion: 3.11.8
 ---
 [Releases](/general/releases) > Moodle 3.11.8 release notes
 
-Release date: Not yet released
+Release date: 11 July 2022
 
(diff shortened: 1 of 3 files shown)


stronk7:
More yet... maybe they can be automatically picked from `data/versions.json` with some helper and we can forget, so every release page will look there to put the release day? Sure @andrewnicols knows about that.

stronk7 (approved):
This looks simple enough for my abilities, hence 👍

andrewnicols:
> More yet... maybe they can be automatically picked from data/versions.json with some helper and we can forget, so every release page will look there to put the release day? Sure @andrewnicols knows about that.

#300 raised to implement this

stronk7:
> https://github.com/moodle/devdocs/pull/300 raised to implement this

Cool, so @mickhawkins only needs to add that to the release template and it will work automagically!

PR #300, opened 2022-07-11
Fetch release note dates from metadata

andrewnicols:
As requested in #298.

This issue introduces a new React component:

```
<ReleaseNoteIntro releaseName="4.0.2" />
```

This will display the:

- release date
- link to the tracker

I've also updated the page description for the meta tag because the import statement and React tags mess with it.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/300"></a>

### DIFF
--- a/.markdownlint-cli2.cjs
+++ b/.markdownlint-cli2.cjs
@@ -164,6 +164,9 @@ config.MD033 = {
         "InvalidExample",
         "ReactPlayer",
         "Link",
+        "ReleaseNoteIntro",
+        "ReleaseDate",
+        "ReleaseVersion",
     ]
 };
 
--- a/general/releases/3.11.md
+++ b/general/releases/3.11.md
@@ -4,11 +4,13 @@ tags:
   - Release notes
   - Moodle 3.11
 sidebar_position: -311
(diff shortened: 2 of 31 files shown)


andrewnicols:
> 1. I imagine those `return null` when a given release is not found means, in practice, print nothing, correct?

Correct - it's known as conditional rendering. You can read more about it here:
https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

> 2. Following on the previous point, could make sense (just asking!) to add the `Release date: to be released` + the link to the list of issues instead of printing nothing?

We can do that :)

We already collect information about versions which aren't yet released - it's something we need to add to https://moodledev.io/general/development/process/release - I notice that it's not a current step for the current list of releases. When we add the new doc page for the next release, we should also be adding the versions.json entry.

I've created #302 - which I thought I had already created but hohum.

> It looks great, I don't quite understand how those helpers work, but I'm sure you do, super Andrew!

sarjona (approved):
Thanks for working on this, Andrew! It's great to keep adding this king of automatisation to the devdocs!! <3

stronk7:
Yay, well done! 🎈

PR #305, opened 2022-07-14
[doc] Migrate component library

sharidas:


general/development/tools/component-library.md
```diff
@@ -0,0 +1,64 @@
...
+  - Themes
+  - Usability
+---
+
+## Component library
+
+This document provides a brief overview of the Moodle Component Library.
+
+## What is the Component Library?
+
```
andrewnicols, line 20: I'd recommend just removing these lines. 
The document already has an h1 title of 'Component Library' so adding an h2 with the same thing is superflous and not great for accessibility.

The 'What is the Component LIbrary?' title is really just an introduction to the content and that's where we should start.

```suggestion
```

general/development/tools/component-library.md
```diff
@@ -0,0 +1,64 @@
...
+The Component Library is written using the same Hugo tooling that the Twitter Bootstrap framework uses. It also includes the Bootstrap libraries for the version of Bootstrap used in Moodle.
+
+The library is built from your console, and you will need to install the same NodeJS dependencies that we use for many of our other development features.
+
+See our [Grunt](https://docs.moodle.org/dev/Grunt) documentation for information on how to install these dependencies.
+
+Once you have installed the NodeJS dependencies you can then use the [grunt](http://gruntjs.com/) tooling to build the library.:
+
+ grunt componentlibrary
```
andrewnicols, line 49: I'd suggest putting this line in a fenced codeblock:

```suggestion
```console title="Building the component library"
grunt componentlibrary
 ```
```

andrewnicols (changes requested):
Just a few really minor things and this can land.

Thanks!

general/development/tools/component-library.md
sharidas: I have removed these lines as per the review in the updated pr.

general/development/tools/component-library.md
sharidas: I have updated the change as per the suggestion. Let me know if this looks ok.

andrewnicols (approved):

PR #312, opened 2022-07-18
[docs] Migrating delegated transactions document

abgreeve:


docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Delegated database transactions
+tags:
+  - core_dml
+  - DML
+  - core
+  - API
+---
+
```
andrewnicols, line 9: I'd suggest adding an introduction (no title please) before the general principles. Nothing too fancy, just something like:

```suggestion

Moodle allows data manipulation to take place within a database transaction, known as a _Delegated transaction_. This allows you to perform CRUD operations, and roll them back if a failure takes place.

```

docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
...
+## General principles
+
+1. No matter of the nested nature, they are really **"delegated transactions"** (so outer levels get the control over inner ones), so we are going to apply that naming to enforce the real concept
```
andrewnicols, line 12: I have no idea what this sentence means. I know this is a document migration, but we're trying to take the opportunity to update documentation at the same time to make it clearer and more meaningful.

docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Delegated database transactions
```
andrewnicols, line 2: It might be better to simplify this to 'Transactions', at least in the sidebar where it's already clear that the topic is a part of the DML API.

If you want the page title to be "Delegated database transactions" still, you can use the `sidebar_label` YAML entry to support this:

```suggestion
title: Delegated database transactions
sidebar_label: Transactions
```

docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
...
+
+7. If, for any reason, the developer needs to catch exceptions when using transactions, it will be mandatory to use it in this way:
+
+    ```php
+    try {
+        $transaction = $DB->start_delegated_transaction();
+        // Perform some $DB stuff
+        $transaction->allow_commit();
+    } catch (Exception $e) {
+        //extra cleanup steps
```
andrewnicols, line 45:
```suggestion
        // Extra cleanup steps.
```

docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
...
+        $transaction->rollback($e); // rethrows exception
```
andrewnicols, line 46: We try to keep code examples relatively short in line length so that they look better on narrower screens.
```suggestion
        // Re-throw exception after commiting.
        $transaction->rollback($e);
```

docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
...
+    }
+    ```
+
+8. In order to be able to keep some parts of code out from top transactions completely, if we know it can lead to problems, we can use:
+
+    ```php
+     $DB->transactions_forbidden(); // Instant check to confirm we aren't using transactions in this point. Will throw exception if transaction is found.
```
andrewnicols, line 53:
```suggestion
     // Check to confirm we aren't using transactions in this point.
     // This will throw exception if transaction is found.
     $DB->transactions_forbidden();
```

docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
...
+    2. Let execution continue, transaction is automatically rolled back in `$DB->dispose()`
+    3. inform with as many details as possible (token, place... whatever)
+3. `$DB->dispose()` will:
+    1. Detect "in-transaction" status
+    2. log error (not possible to honour settings!)
+    3. Properly perform the full DB rollback
+
+## Related tasks
+
+1. Move this to MoodleDocs and publicly in Dev Forum
```
andrewnicols, line 76:
```suggestion
```

docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
...
+2. Any "footer" (meaning some place before ending `<html>` output) will:
+    1. Detect "in-transaction" status
+    2. Let execution continue, transaction is automatically rolled back in `$DB->dispose()`
+    3. inform with as many details as possible (token, place... whatever)
+3. `$DB->dispose()` will:
+    1. Detect "in-transaction" status
+    2. log error (not possible to honour settings!)
+    3. Properly perform the full DB rollback
+
+## Related tasks
```
andrewnicols, line 74: Are these tasks still relevant? Are they from when this was a spec?

Ultimately we do not make great use of transactions in Moodle - and I don't know whether we _should_ or not!

1) is obviously done or moot now because txns have been around for so long
2) hopefully we have tests for this!?
3) this probably needs to be made clearer in various places
4)  ??
5) Not really sure if this belongs in this doc

docs/apis/core/dml/delegated-transactions.md
```diff
@@ -0,0 +1,80 @@
...
+2. Code will **never** rely on rollback happening. It is only a measure to reduce (not to eliminate) DB garbled information
+3. Any problem related with the use of transactions (unfinished ones, unbalanced, finished twice...) will end **always** with one transaction_exception that will **always** end performing one DB rollback
+4. If one transaction (at any level) has been marked for rollback() there won't be any method to change it. Finally Moodle will perform the DB rollback
+5. If one transaction (at any level) has been marked for commit() it will be possible to change that status to rollback() in any outer level
+6. It will be **optional** to catch exceptions when using transactions, but if they are caught, then it is mandatory to mark the transaction for rollback()
+7. Any explicit rollback() call will pass the exception originating it, i.e. rollback($exception) to be re-thrown
+8. API will work in a natural and similar fashion to current recordsets one
+9. We will start **recommending** to use InnoDB for 2.0 in environmental tests. Possibly, at some point in the future that will become a **must**. But **it is not** for now (be warned!)
```
andrewnicols, line 20: Most of these could do with a little re-wording to clarify their meaning.

We've also been generally trying to avoid the use of `e.g.` and `i.e.` preferring "for example" and "that is" a lot of the time - only because most style guides (Microsoft, MDN, and others) suggest that it isn't as clear for non-native speakers.

andrewnicols (changes requested):
Hi Adrian,

Thanks for looking at this migration. If you're able to make a few change to clarify things that would be great, otherwise we can accept this now and raise an issue to fix these later.

Some of the suggestions are easy, and others require rewording some of the existing wording because it's just not very clear. If you want to make the easy ones, that's fine too - just let us know :)

docs/apis/core/dml/delegated-transactions.md
andrewnicols: Notes from the MS style guide:
- i.e. - https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/i/ie
- e.g. - https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/eg

MDN guidance (see note below table): https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Writing_style_guide#latin_abbreviations

andrewnicols (approved):

docs/apis/core/dml/delegated-transactions.md
abgreeve: I've added the text that you have here.

docs/apis/core/dml/delegated-transactions.md
abgreeve: I've removed this line. I don't know what it means either.

docs/apis/core/dml/delegated-transactions.md
abgreeve: I've just changed the main title with out adding the sidebar label. Happy to change this to a side bar label if that is preferred.

docs/apis/core/dml/delegated-transactions.md
abgreeve: Updated to be like this

docs/apis/core/dml/delegated-transactions.md
abgreeve: This has been updated as well.

docs/apis/core/dml/delegated-transactions.md
abgreeve: Changed to this.

docs/apis/core/dml/delegated-transactions.md
abgreeve: Removed.

docs/apis/core/dml/delegated-transactions.md
abgreeve: I've removed all of these. This has been around long enough that it is either documented or it doesn't make sense.

docs/apis/core/dml/delegated-transactions.md
abgreeve: I have gone an reworded these a bit. It probably needs another review. 
A few comments about this section:
- Point 2. It seems that the author thought this to be very important (it's mentioned in two different areas of the document) but I don't find the comment useful at all.
- Point 8 does not seem useful at all.
- Point 9 I'm not sure what the status of testing is using transactions. Perhaps you can give some guidance here on what this means?

abgreeve:
I have a second commit here. Please let me know if you would like this squashed.

Issue #248, opened 2022-06-23
High priority migrations


andrewnicols:
### What do we need to do?

The following are high priority migrations.

If you are able to move any of these files, please help us out by doing so.

If you are working on one of these, please add a link to the issue and check it

[ ] https://docs.moodle.org/dev/Activity_completion_API
[ ] https://docs.moodle.org/dev/Admin_settings
[ ] https://docs.moodle.org/dev/Advanced_grading_API
[ ] https://docs.moodle.org/dev/Availability_API to /docs/apis/subsystems/availability/index.md
[ ] https://docs.moodle.org/dev/Backup_API to /docs/apis/subsystems/backup/index.md
[ ] https://docs.moodle.org/dev/Cache_API to /docs/apis/subsystems/muc/index.md
[ ] https://docs.moodle.org/dev/Cache_API_-_Quick_reference to /docs/guides/muc/index.md ?
[ ] https://docs.moodle.org/dev/Calendar_API
[ ] https://docs.moodle.org/dev/Check_API 
[ ] https://docs.moodle.org/dev/Comment_API to /docs/apis/subsystems/comment/index.md
[ ] https://docs.moodle.org/dev/Competency_API
[ ] https://docs.moodle.org/dev/Conditional_activities_API
[ ] https://docs.moodle.org/dev/Custom_fields_API
[ ] https://docs.moodle.org/dev/Editor_API
[ ] https://docs.moodle.org/dev/Events_API
[ ] https://docs.moodle.org/dev/Experience_API_(xAPI)
[ ] https://docs.moodle.org/dev/External_functions_API
[ ] https://docs.moodle.org/dev/Gradebook_API
[ ] https://docs.moodle.org/dev/Lock_API to /docs/apis/subsystems/lock/index.md
[ ] https://docs.moodle.org/dev/Logging_2
[ ] https://docs.moodle.org/dev/Media_players
[ ] https://docs.moodle.org/dev/Message_API to /docs/apis/subsystems/message/index.md ?
[ ] https://docs.moodle.org/dev/My_profile_API
[ ] https://docs.moodle.org/dev/OAuth_2_API
[ ] https://docs.moodle.org/dev/Payment_API
[ ] https://docs.moodle.org/dev/Portfolio_API
[ ] https://docs.moodle.org/dev/Preference_API
[ ] https://docs.moodle.org/dev/Privacy_API
[ ] https://docs.moodle.org/dev/Question_API
[...]

PR #314, opened 2022-07-18
[docs] Migrate Backup and Restore API pages.

abgreeve:


data/component-spelling.txt
```diff
@@ -23,6 +23,7 @@ cohort
 comment
 competency
 completion
+configdata
```
andrewnicols, line 26: This file is an automatically generated list of subsystems and plugintypes (it will eventually also include the standard moodle plugin list too).
Manual changes to this will get automatically removed on Sunday.

You probably want to put these in project-words.txt instead.

andrewnicols (changes requested):
Just the changes to the component-spelling.txt in the wrong place. Otherwise this is good to go.

data/component-spelling.txt
abgreeve: Thanks for pointing me in the right direction here. I have moved these to project-words.txt.

andrewnicols (approved):

stronk7:
Hi, nice addition, well done!

Just a side note, for your consideration... I had this here in my TODOs... 

Link (not sure it works, first time I try it ever): [Moodle Developer Resources chat](https://matrix.to/#/!AJVqqurwTqtyNlgxOd:moodle.com/$qGQAbe34LxDqVLR7LlLGuLcuOhTDVA_fFWnoXNCp6Zo?via=moodle.com&via=matrix.org&via=k3s.ems.host).

Original comment:

> Q: Looking to https://github.com/moodle/devdocs/issues/248 I had plans to give it a try with https://docs.moodle.org/dev/Backup_API (one of the elements in the list).
> 
> Problem (conceptually) that I'm facing is that, from when I did look for related dev docs recently (marking some of the pages like "not to migrate). I came with this list about pages to migrate:
>
> h4. 100% sure:
>- https://docs.moodle.org/dev/Backup_API
>- https://docs.moodle.org/dev/Backup_2.0_for_developers
>- https://docs.moodle.org/dev/Restore_API
>- https://docs.moodle.org/dev/Restore_2.0_for_developers
> h4. Not so sure:
>- https://docs.moodle.org/dev/Backup_2.0_theme_data
>- https://docs.moodle.org/dev/Backup_2.0_course_report_data
> So, the question is... what should go to the "canonical" /docs/apis/subsystems/backup/[index.md](http://index.md/) new page. Just an intro and then link to others... or maybe everything together? Or, how do we split backup and restore?
>
> I mean, I'm not sure about which the structure should be.

Just sharing here in case we want to do anything with those other pages (migrate, obsolete... whatever). Not the original question about locations.

Ciao :-)

andrewnicols:
Thanks Eloy,

I'm getting to be of the opinion that perfect is the enemy of good - that is that we're better off at this time to get as many docs moved over as quickly as possible and then to refine and improve them later.

We definitely want to move some/all of those other pages and/or merge them into the recently migrated ones. I'm also of the opinion that we should be writing a developer _guide_ as well.

I see the two being slightly different:
- docs/apis/subsystems - information on subsystems, how to use them, and key points
- docs/apis/plugintypes - information for those developing a plugintype, and key points
- docs/guides - walkthroughs on how to do a common action or use a common API from start to end - useful for more detailed items like the JavaScript guide, or the Templating guide, or other more general concepts.

PR #326, opened 2022-07-27
[Docs] fixed minor typos in the sample implementation of mod_myplugin_pluginfile

wannerdev:
Thanks for the sample implementation :)
fixed some minor typos

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/326"></a>

### DIFF
--- a/docs/apis/subsystems/files/index.md
+++ b/docs/apis/subsystems/files/index.md
@@ -193,7 +193,7 @@ function mod_myplugin_pluginfile(
         // false here.
     }
 
-    // For a plugin which does not specify the itemid, you may want to use:
+    // For a plugin which does not specify the itemid, you may want to use the following to keep your code consistent:
     $itemid = null; //to make your code more consistent.
 
     // Extract the filename / filepath from the $args array.
--- a/docs/apis/subsystems/files/index.md
+++ b/docs/apis/subsystems/files/index.md
@@ -143,7 +143,7 @@ function [component_name]_pluginfile(
 function mod_myplugin_pluginfile(
     $course,
     $cm,
-    $context.
+    $context,
     string $filearea,
     array $args,
     bool $forcedownload,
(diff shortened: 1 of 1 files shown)


docs/apis/subsystems/files/index.md
```diff
@@ -215,7 +215,8 @@ function mod_myplugin_pluginfile(
     }
 
     // We can now send the file back to the browser - in this case with a cache lifetime of 1 day and no filtering.
-    send_stored_file($file, DAY_SECS, 0, $forcedownload, $options);
+    $daySecs = 60*60*24;
```
andrewnicols, line 218: This is incorrect. THe existing line is correct. Please use the `DAY_SECS` constant rather than creating new variables (with incorrect coding naming conventinos)

docs/apis/subsystems/files/index.md
```diff
@@ -194,7 +194,7 @@ function mod_myplugin_pluginfile(
     }
 
     // For a plugin which does not specify the itemid, you may want to use:
```
andrewnicols, line 196:
```suggestion
    // For a plugin which does not specify the itemid, you may want to use the following to keep your code consistent:
```

docs/apis/subsystems/files/index.md
```diff
@@ -194,7 +194,7 @@ function mod_myplugin_pluginfile(
...
-    $itemid = null to make your code more consistent.
+    $itemid = null; //to make your code more consistent.
```
andrewnicols, line 197: I'd suggest leaving this as a comment as it is an _alternative_ to the code above.
```suggestion
    // $itemid = null;
```

andrewnicols (changes requested):
THis looks good, but a few minor changes:
* The itemid should actually be a comment as it is an alternative to the code above it
* The change to remove use of the DAY_SECS constant is wrong. We should use the time constants wherever possible

Finally, it would be good if you could amend your commit message as 'Update index.md' does not describe the change, or what the change relates to.

Thanks

docs/apis/subsystems/files/index.md
wannerdev: Hm interesting i can't access this constant.

andrewnicols: Ah - that's because it is DAYSECS and not DAY_SECS :)
Could you update your patch to fix the naming and then we hsould be all good :)

wannerdev:
I did look at the last commit

andrewnicols (approved):

PR #334, opened 2022-08-09
[docs] Fixing an error on deprecating a parameter doc

aanabit:
Add a mention to corresponding upgrade.txt documenting the deprecated PARAMETER should not be used any more.

instead of

Add a mention to corresponding upgrade.txt documenting the deprecated method should not be used any more.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/334"></a>


sarjona (approved):
Thanks Amaia for raising and fixing this :-) Well done!

PR #352, opened 2022-09-08
Remove step from minors, it only applies to majors

stronk7:
Agreed in chat, this step is only applicable to majors, when a new branch arrives.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/352"></a>


andrewnicols (approved):

junpataleta (commented):
Looks spot on. Thanks!

PR #354, opened 2022-09-14
Fix incorrect slug

stronk7:
I imagine this is the cause for the old docs pointing to incorrect URL in the new docs. AKA:

https://docs.moodle.org/dev/Moodle_and_PHP

is pointing to:

https://moodledev.io/general/development/php

and it should point to:

https://moodledev.io/general/development/policies/php

It seems that in the last move of that page, the information was modified only partially.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/354"></a>


stronk7:
Note it's just a supposition, so tell me if I'm wrong, heh!

andrewnicols (approved):

PR #355, opened 2022-09-15
[chore] Stop some workflows running on all clones

### LINKED ISSUE #299: [chore] Prevent various GHA workflows to be run on forks
### What do we need to do?

Recently I've detected that various GHA workflows are being executed in my fork repo:

https://github.com/stronk7/devdocs/actions

Those only should happen on the "central/upstream" one.

I imagine that can be done with some restriction/condition of the type `if: github.repository_owner == 'moodle'` (we have [used it in other repos](https://github.com/moodlehq/moodle-plugin-ci/blob/master/.github/workflows/release.yml#L9-L10)).

andrewnicols:
Fixes #299

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/355"></a>

### DIFF
--- a/.github/workflows/fetch-obsolete-pages.yml
+++ b/.github/workflows/fetch-obsolete-pages.yml
@@ -15,6 +15,9 @@ on:
 
 jobs:
   run:
+    # Only run on the canonical repository.
+    if: github.repository == 'moodle/devdocs'
+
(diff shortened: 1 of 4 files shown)


stronk7 (approved):
This looks perfect! Approved!

PR #358, opened 2022-09-15
Migrate Activity completion API docs

andrewnicols:
This includes two commits:
* migrate the existing docs in a format as close to original as possible
* update docs to suck less

I have not attempted to make these docs perfect but I think we need to just get docs migrated over ASAP now and improve quality later.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/358"></a>


andrewnicols:
Self-merging after one week.

PR #385, opened 2022-10-10
[docs] Migrate License documentation

andrewnicols:
This includes a redirect for /license

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/385"></a>

### DIFF
--- a/data/migratedPages.yml
+++ b/data/migratedPages.yml
@@ -207,6 +207,9 @@ Hardening_new_Roles_system:
 Integration_Review:
 - filePath: "/general/development/process/integration/index.md"
   slug: "/general/development/process/integration"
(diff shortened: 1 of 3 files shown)


stronk7 (approved):
Yay!

Only point I'm not sure is which one should be the main page (general/license or license) and which the redirect... but I imagine it's not really important and keeping the main under general is ok.

stronk7:
Merged, thanks!

andrewnicols:
With our current configuration, all of our docs must be in either the
general or docs folders. We can have top level redirects to them, or we can
have static pages at other points, but I don’t think we can have a page
which is a part of the main documentation there.

PR #393, opened 2022-10-18
[docs] Document the new TinyMCE plugin API

andrewnicols:
MDL-75966


andrewnicols:
Documentation for MDL-75071

docs/apis/plugintypes/tinymce/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: TinyMCE Editor plugins
+title: Legacy TinyMCE Editor plugins
```
sarjona, line 2: As we've left only a single reference to "TinyMCE Editor plugins" in docs/apis/plugintypes, I would suggest moving this Legacy to a subfolder (having "TinyMCE Editor Plugins" as parent).

docs/apis/plugintypes/tiny/index.md
```diff
@@ -0,0 +1,586 @@
...
+  - editor_tiny
+  - tiny_
+  - Editor
+  - Text editor
+  - HTML editor
+  - WYSIWYG
+  - TinyMCE
+---
+
+Moodle includes the TinyMCE text editor as standard from Moodle 4.1, and it can be installed from the plugins database for Moodle versions 3.11, and 4.0.
```
sarjona, line 13: I guess for now it hasn't been uploaded to the plugins database yet, right? (at least, I can't find it)

docs/apis/plugintypes/tiny/index.md
```diff
@@ -0,0 +1,586 @@
+---
+title: TinyMCE Editor Plugins
+tags:
+  - editor_tiny
+  - tiny_
```
sarjona, line 5: What's the point of adding "_" at the end of this tag?

docs/apis/plugintypes/tiny/index.md
```diff
@@ -0,0 +1,586 @@
...
+
+The `editor_tiny` editor supports the inclusion of subplugins, which have the namespace `tiny_[pluginname]`.
+
+import {
+    Lang,
+} from '../../_files';
+
+## File structure
+
+Course custom field plugins are located in the `/lib/editor/tiny/plugins` directory. A plugin should not include any custom files outside of it's own plugin folder.
```
sarjona, line 23:
```suggestion
TinyMCE Editor plugins are located in the `/lib/editor/tiny/plugins` directory. A plugin should not include any custom files outside of its own plugin folder.
```

sarjona (changes requested):
Hi @andrewnicols!
Thanks for working on this (the TinyMCE plugin and also its documentation). It looks great and, as you'll see, only a few minor points have been raised :-)

docs/apis/plugintypes/tiny/index.md
andrewnicols: Correct - I don't plan to upload it until about a week after it's been integrated to core.

docs/apis/plugintypes/tiny/index.md
andrewnicols: It's the subplugin prefix. Any alternative suggestion?

docs/apis/plugintypes/tiny/index.md
andrewnicols: Good catch - fixed thanks.

docs/apis/plugintypes/tiny/index.md
sarjona: I was thinking to leave only "tiny". If you look at how tags work, you'll see the "_" is ignored: https://deploy-preview-393--moodledevdocs.netlify.app/docs/tags/tiny (so I think displaying "tiny" instead of "tiny_" would be clearer).

andrewnicols: Thanks - I've updated that.

PR #398, opened 2022-10-24
Bump the priority of non-bugs related with the release

stronk7:
Add a point to -4w before release (1w after freeze) about to bump the priority of all the minor/trivial non-bug issues to major.

This includes both must-fix issues and issues already under IR/CLR which unique affected version is the one to be released. (the query needs to be amended on every release, it's noted in the new step description).

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/398"></a>


junpataleta (approved):
Thanks, Eloy. Looks good to me.

PR #427, opened 2022-11-13
[repo] Clear the content of migrated pages on wikimedia

andrewnicols:
When marking a page as migrated in Wikimedia we should clear the page content. It's time we started to clear out the old site!


sarjona (approved):
Hi Andrew!
The patch looks good! I like the idea of removing the content (it will help people to move to the new moodledev docs). Well done! :-)

PR #428, opened 2022-11-13
[repo] Upgrade to Docusaurus 2.2.0

andrewnicols:


andrewnicols:
Self-merging because to avoid this getting out-of-date.

PR #432, opened 2022-11-14
Update scssphp in thirdpartylibs

davewoloszyn:


sarjona (approved):
https://tracker.moodle.org/browse/MDL-75476 is the issue where this library was upgraded.
Thanks for working on a patch for this, @davewoloszyn!

PR #433, opened 2022-11-14
Update PHP CSS Parser in thirdpartylibs

davewoloszyn:


sarjona (approved):
This library was upgraded in https://tracker.moodle.org/browse/MDL-75475
@davewoloszyn thanks for working on a patch for updating the devdocs page :-)

PR #434, opened 2022-11-14
Update Mustache in thirdpartylibs

davewoloszyn:


sarjona (approved):
https://tracker.moodle.org/browse/MDL-75473 is the issue where this library was upgraded.
@davewoloszyn thanks for updating the version for it in the devdocs page.

PR #448, opened 2022-11-22
[docs] Add links to PHP 8.2/8.3 epics

andrewnicols:
This change also moves the list of unsupported versions to the bottom of the page.


stronk7 (approved):
Ah, wow, you've created the 8.2 and 8.3 epics in advance (unexpected, hope all the details and TODOs have been cloned there, so we don't forget about them.

This is ok for me, +1.

PR #449, opened 2022-11-23
Core availability API: code example of setting date availability

marcusgreen:


docs/apis/subsystems/availability/index.md
```diff
@@ -134,6 +134,31 @@ $DB->set_field(
...
+The following code can be used to programmatically set start and end date restrictions.
+
+```php
+use \core_availability\tree;
+
+$dates = [];
+$dates[] = \availability_date\condition::get_json(">=", $availability['start']);
+$dates[] = \availability_date\condition::get_json("<", $availability['end']);
+
+$showc = [true, true];
```
andrewnicols, line 146: What, prey-tell, does this magic _mean_? Is it docuemnted anywhere?

catalyst-marcus-green: If you are referring to $showc = [true, true];
I had exactly the same question. Line 160 describes it but a // comment above it might make it clearer. It was wildly frustrating to get this to work, and at one point I had a big long text string until someone showed me the example above and this is where I came up with the more programmery version of it.

andrewnicols: So are the array key meanings matched with the dates? That is:
```
$showc = [
    true, // start
    true, //end
];
```

catalyst-marcus-green: Yes, it means that both restrictions will be set with the eye open/closed (in the interface), which controls if the activity on a course will be either greyed out or not shown at all. This is checked somewhere in the code, but I was not able to find where that was over the weekend. Somewhere it does a count of the items, and compares with the count of the showc arrays

catalyst-marcus-green: This is the original code when I had not understood how to do this properly
https://github.com/marcusgreen/moodle-tool_driprelease/blob/8d24bd8059ce8e539b5df442449a57ffe4cd7b1e/lib.php#L194

And what I moved to and based this example on
https://github.com/marcusgreen/moodle-tool_driprelease/blob/d0fd4815834095da9f9b3c1de9eeecba58995af8/lib.php#L198

andrewnicols (approved):

PR #451, opened 2022-11-25
[docs] Complete 4.1 release notes highlights

mickhawkins:
Final release notes for LMS 4.1 major release.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/451"></a>

### DIFF
--- a/general/releases/4.1.md
+++ b/general/releases/4.1.md
@@ -56,10 +56,257 @@ Mobile:
 
 For the best experience and optimum security, we recommend that you keep your browser up to date.
 
-## Major features
+## Major UX improvements
 
-Coming soon.
+### Gradebook
+<!-- cspell:disable -->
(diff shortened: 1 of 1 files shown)


stronk7:
Hi,

as commented elsewhere... the only detail that called my attention is the one about Safari 7 (that we keep dragging over all the release notes... it's from 9-10 years ago...

I mean, we don't comment about Chrome 25 (that is from the same year) and surely doesn't work ok either.

In any case, that detail doesn't stop this to be published IMO. So I'm going to proceed now. Pending:

1. Someone to review if there is anything missing (epics, issues lists).
2. Everybody review if there is anything worth being added.
3. The "Safari" detail, for your consideration.
4. Add vids or links to new features...

Ciao :-)

stronk7 (approved):

PR #455, opened 2022-11-29
Replace Box/Spout with OpenSpout

meirzamoodle:
https://tracker.moodle.org/browse/MDL-76219


sarjona:
@meirzamoodle Now that https://github.com/moodle/devdocs/pull/497 has been integrated, this patch needs to be reviewed because, now that versions are automatically upgraded from thirdpartylibs.xml, the only required change is to add an entry to spout in the legacy libs.

PR #459, opened 2022-12-02
Bump dependencies

andrewnicols:
This patch is a maintainance task to bump dependencies.

I've included addin gthe yarn interactive tools plugin to make this easier. This can be run using:
```
yarn interactive-upgrade
```

From there you can choose which dependencies to upgrade, and to which versions.

I've also had to fix a number of existing lint violations which have crept in as this change bumps lint version dependencies.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/459"></a>


sarjona (approved):
Hi @andrewnicols!
Thanks for taking care of this and bumping some versions :-) 
The patch is working properly, so I'm happy to approve and merge it!

PR #460, opened 2022-12-02
[repo] Fix broken sidebar build

andrewnicols:
This addresses two issues:
1. Broken sidebar build on Windows due to use of / instead of path.resolve or path.join; and
2. A warning for a new major release which does not have any minor releases yet.

Unfortunately I'm unable to test the Windows case, but the following warning is no longer present when building as a result of this change:
```sh
[WARNING] No docs found in "releases/4.1": can't auto-generate a sidebar.
```

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/460"></a>


sarjona (approved):
Hi Andrew!
Thanks for reporting and working on this. Although I haven't been able to confirm if the error is fixed in Windows, I can see the warning you mentioned has disappeared for me too :-)

PR #473, opened 2022-12-09
Release Moodle 4.1 docs

andrewnicols:
This is our first version fork.

It was generated by running:

```
yarn run docusaurus docs:version 4.1
```

There were a few minor pages to modify:
- intro.md in both 4.1 and master docs (pointing to 4.0)
- navbar configuration to uncomment the version dropdown

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/473"></a>


stronk7 (approved):
Hi,

I've been playing a little bit with the auto-deployed version and everything seems to work ok. Using the versions drop-down is possible to switch between versioned docs (apis...) easily. So I think this PR per se is ok and can be accepted to get the first experiences with the multi-versioned docs.

Only points that I've found unnatural (or weird) are also related with the versions drop-down behaviour itself. Here there are a few points where I think that we could improve, all around that drop-down menu:

1. Does the drop down need some label? Maybe not, but right now it looks like (to me, this is a 100% subjective feeling) it's floating in the middle of nowhere.
2. If you hover the menu (or click on the "triangle") it correctly shows you the available options (master, 4.1 and legacy). Then can select any of them, all ok. But, if you click on the "name" of the current version, then you're redirected to the "base" document in that version. IMO, that's completely unexpected. The "name" should behave the very same than the "triangle" or become 100% disabled.
3. Related with the above... the existence of the versions drop-down in non-versioned docs (say intro, releases, processes...) is really miss-leading (there aren't versions for those docs) and trying to switch of version (via "name" like in the previous point, or via "hover" or "triangle") leads you to the unexpected "base" page. Maybe we should hide or disable the drop-down completely in non-versioned pages?

Just those 1-2-3 strange things I found, for your consideration. As said, other than that... let's approve this now, yay!

Ciao :-)

andrewnicols:
I've raised #477, #478, #479 to look at those suggestions.

I agree with all of them. I'll try and find a good way of doing them.

stronk7:
Cool, thanks!

PR #480, opened 2022-12-12
[repo] Only show version dropdown in versioned docs

### LINKED ISSUE #478: [chore] Only show version selection dropdown on versioned docs
### What do we need to do?

A suggestion from Eloy re versioned documentation:

> Related with the above... the existence of the versions drop-down in non-versioned docs (say intro, releases, processes...) is really miss-leading (there aren't versions for those docs) and trying to switch of version (via "name" like in the previous point, or via "hover" or "triangle") leads you to the unexpected "base" page. Maybe we should hide or disable the drop-down completely in non-versioned pages?

andrewnicols:
Fixes #478 

This wraps the standard `DocsVersionDropdownNavbarItem` component from Docusaurus and uses the active document context information to determine whether the current document _has_ any version.

If no version is present, then the code shortcircuits and just returns an empty string.
If there is a version, then the version dowpdown component is called.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/480"></a>

### DIFF
--- a/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js
+++ b/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.js
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) Moodle Pty Ltd.
+ *
+ * Moodle is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


sarjona (approved):
Hi @andrewnicols!
Thanks for working on this issue and removing the Versions selector from the unversioned sections! The patch looks good and is working as expected :-)

PR #491, opened 2022-12-20
Migrate ws docs

andrewnicols:
This is a first attempt at migrating _some_ of the web service documentation, and improving it.

It's not perfect, but we have to start somewhere. Some of the old documentation was project documentation talking more about what we were thinking of implementing, and very difficult to follow. There was a lot of duplication.

There is still a lot of duplication but it is significantly better than it was.

Other docs will be moved over in different chunks. I can only do so many of these are a time.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/491"></a>


andrewnicols:
Note: This needs backporting to the 4.1 docs.
I'll hold off on doing so until it's been reviewed. If all is good, I'll backport it and it can be merged.

PR #492, opened 2022-12-21
Moodle App ver sion 4.1 and minor improvements

jleyva:


andrewnicols (approved):
For future reference, we've added a remark plugin to turn tracker issues matching MDL, MOBILE, and MDLSITE (and possibly others) into links. You can just put MOBILE-12345 and it will become a link in the compiled markdown.

PR #493, opened 2022-12-21
[docs] Add proposed wording for MDLSITE-4776 short arrays

andrewnicols:


sarjona (approved):
In https://tracker.moodle.org/browse/MDLSITE-4776 has been agreed to update the policy for using Short Array Syntax.
Thanks for preparing this patch, Andrew! It looks good so I'm happy to merge it!

PR #497, opened 2022-12-30
[repo] Generate Third-party libraries from repository

### LINKED ISSUE #93: [chore] Update third-party libraries  from thirdpartylibs.xml file
### What do we need to do?

While reviewing https://github.com/moodle/devdocs/pull/87, Andrew suggested that with the libraries we should:

- update the existing thirdpartylibs.xml files in core and add:
  - whether customisations have been made (`<customised />` if true)
  - the URL source of the library
  - the copyright owner
- create a script to parse the various thirdpartylibs files and turn them into markdown
- turn that into a Github action run weekly

andrewnicols:
This change includes:
- a new libraries tool
- a command on the tool (fetch) which will parse the thirdpartylibs.xml files located in the .moodle directory and then generate a libraries.json file from this
- a command on the tool (generate) which will take the libraries.json file and a legacy-libraries.json file, and then generate a static markdown file
- a GitHub action to run these commands once per week

Fixes #93

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/497"></a>


data/legacy-libraries.json
```diff
@@ -0,0 +1,232 @@
+[
```
sarjona, line 1: There are 2 missing legacy libraries that should also be included in this file:

- MathJax: https://moodledev.io/general/community/credits/thirdpartylibs#mathjax
- mp3player: https://moodledev.io/general/community/credits/thirdpartylibs#mp3player

scripts/librarian.mjs
```diff
@@ -0,0 +1,200 @@
...
+- **Version**: ${libraryVersion}${renderCustomisedNotice(libraryCustomised)}
+- **License**:  ${libraryLicense} ${libraryLicenseVersion}
+${renderLibraryUrl(libraryUrl)}
+${renderCopyrightHolders(libraryCopyrightHolders)}
+`.replaceAll(/ *$/gm, '');
+};
+
+const generateCredits = async () => {
+    const generateLibraryContentForFile = async (filePath, isOutdated) => JSON.parse(await readFile(filePath, 'utf-8'))
+        .sort((a, b) => a.location.localeCompare(b.location))
```
sarjona, line 164: I would suggest keeping the alphabetical order (by library name)... IMHO, it makes it easier to find them.

sarjona (changes requested):
Hi @andrewnicols!
Thanks a lot for working on this nice improvement! It will save us a lot of time and will also help us to keep this page upgraded (reviewing the patch I've realised we were missing a few libraries :-( ).
Well done!! :-)

data/legacy-libraries.json
andrewnicols: Done

scripts/librarian.mjs
andrewnicols: In theory, sure...but right now we have some cases where there isn't a name.

`theme/boost/thirdpartylibs.xml` is missing a name for `amd/src/bootstrap/button.js` for example.

Until the xml is validated against an XSL I think we have to go with the one field which must be there (location) because other automatinos already rely on it within Moodle.

sarjona: Ohhhh, fair enough!

sarjona (approved):
Thanks Andrew!! I think the patch is ready to be merged.
Let's see if we can add the XLS in Moodle to guarantee all the thirdpartylibs.xml files have name and then, we'll be able to change the default sorting :-)

PR #513, opened 2023-01-10
Accessibility docs page tweaks

junpataleta:
- Make the wording more generic and not just pertaining to a single occasion of the accessibility audit.
- Remove the accreditation badges (especially since they are currently expired) and replace them with links to the accreditation pages.
- List down the dates when WCAG 2.1 Level AA accreditation was issued on the given Moodle versions on a table.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/513"></a>

### DIFF
--- a/general/development/policies/accessibility.md
+++ b/general/development/policies/accessibility.md
@@ -10,36 +10,28 @@ Moodle is designed to provide equal functionality and information to all people.
 
 ## Accessibility conformance
 
-As part of our ongoing commitment to provide an accessible platform and to continuously improve compliance with WCAG 2.1 Level AA, Moodle has been working with an external auditor to review the platform since January 2020.
+As part of our ongoing commitment to provide an accessible platform and to continuously improve compliance with WCAG 2.1 Level AA, Moodle has been working with an external auditor to review the platform since the development of Moodle 3.9 in January 2020.
 
-Key pages within Moodle, which are representative of its overall accessibility and functionality, are audited using both automated tools and user journey testing. The WCAG 2.1 Level A and Level AA issues raised from this audit were all addressed in Moodle 3.10 and Moodle 3.9.3 (both released on November 9th 2020). Many of these [accessibility improvements](https://tracker.moodle.org/browse/MDL-67688) where also back-ported to Moodle 3.7.5+ and Moodle 3.8.2+, where possible.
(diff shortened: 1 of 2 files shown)


andrewnicols (approved):
Is it worth adding a link to the relevant issue for each of the WCAG reviews?

Also is it worth noting that the accreditation does expire, but that we still achieved compliance in the first place and that we do not recertify versions after their initial certification. That seems to be the number two recurring question after "When is Version X going to be certified?"

I'll let you self-merge this if you want to now, or you can add a patch with any of the above here or in a new issue. Up to you.

junpataleta:
Thanks, @andrewnicols. I added a [fixup commit](https://github.com/moodle/devdocs/pull/513/commits/a5d2bf96859d42c421b2115fb2b1d58f32e3331b) where:
- The epic links for the accessibility audits are added on the table for Moodle LMS
- I added a note about the expiry of the accreditation. Though I am not sure if I worded it appropriately or whether it can be further improved/made clear. Please feel free to suggest improvements. 

Marie will also be checking this out sometime soon and am looking for her feedback too.

Cheers!

general/development/policies/accessibility.md
```diff
@@ -16,18 +16,18 @@ Key pages within Moodle, which are representative of its overall accessibility a
 
 :::info Moodle's accreditations
 
-Both Moodle LMS, and the Moodle Mobile App hold a WCAG 2.1 AA accreditation based on the successful resolution of issues identified during audits.
+Both Moodle LMS and the Moodle Mobile App hold a WCAG 2.1 AA accreditation based on the successful resolution of issues identified during audits.
 
-These accreditations are renewed each year.
+The accreditations issued do expire after a year. Though it does not mean that the Moodle versions that were accredited are no longer accessible nor they are no longer supported for accessibility. It just means that the process for the accreditation of the latest supported Moodle version(s) are still in progress or pending.
```
andrewnicols, line 21: I'd suggest something like t his maybe:

```suggestion
Accreditations expire after one year, but this does not mean that the accreditation is invalid or that the Moodle version is no longer accessible. An expired accreditation just means that the process for re-accrediting the latest supported version or versions of Moodle is currently in progress.
```

Remove the distinction between "in progress" and "pending", and put it into a slightly more complete and full sentence.

junpataleta: Thanks, Andrew! I have applied your suggestion, amended the fixup commit and squashed it to the original commit.

PR #517, opened 2023-01-13
[docs] Update for 4.1.1 minor release

mickhawkins:


mattporritt (approved):
This all looks good.
The release graph image is correct and matches what we put in the headsup
The versions are correct
The release note file matches what it should be
The issues in the release notes look correct (I did the initial tracker in the tracker)

PR #518, opened 2023-01-17
docs: Fix dates for 3.6 release

andrewnicols:


mattporritt (approved):
Yep, that should fix it

PR #522, opened 2023-01-18
Address an issue with multiple MDL-s on the same line not being converted to tracker links

andrewnicols:
The AST processor needed to be told the index of the next node to search in the parent.

Previously we were skipping and stopping processing at that point (`return null == 'skip'`).

When we make a modification to the AST, we must also then return the next index to search - that is the index after the changes.

In this case the next index is previous index + 2. We insert two new items onto the tree:
- the link
- the remainder of the line

We need to process the remainder of the line.

This changeset adds three new unit tests, two of which were previously failing due to this issue.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/522"></a>

### DIFF
--- a/src/remark/trackerLinks.js
+++ b/src/remark/trackerLinks.js
@@ -26,8 +26,6 @@ const projects = [
     'CONTRIB',
 ];
 
-const expression = new RegExp(`(?<issueNumber>(${projects.join('|')})\\-\\d+)`, 'g');
-
 /**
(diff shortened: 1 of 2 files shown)


sarjona (approved):
Thanks for fixing this issue. It's always a pleasure to review your code and learn from you!! <3
The patch looks good (and it's great you also covered it with some unit tests). Well done!

PR #536, opened 2023-01-30
4.1 iTeam retrospective amendments to the release process

junpataleta:
* The continuous integration period starts 7 weeks before the release
* Code freeze starts 6 weeks before the release
* QA cycle starts 5 weeks before the release


stronk7 (approved):
Thanks Jun,

I've checked everything twice and it seems correct, with all the related tasks moved perfectly 1 week up.

Tangentially related, I've edited the description in some of the CI jobs (queue manager jobs) to match the new 7 weeks of continuous with freeze happening on -6. I don't think we have that information elsewhere, apart from calendars.

Good one!

PR #538, opened 2023-01-30
include \core\check\result

ewallah:


andrewnicols (approved):
Good catch. Thank you :)

PR #544, opened 2023-02-02
Add notes about `$required` parameter validation

junpataleta:


andrewnicols (approved):
Good clean update. Love it!

PR #553, opened 2023-02-14
[docs] Add the missing $PERF variable in the list of global variables

jboulen:


andrewnicols (approved):

andrewnicols (approved):

andrewnicols:
I don't believe that this is intended to be an exhaustive list, but it's good to have it more complete.

Others still missing include `$ADMIN`.

PR #555, opened 2023-02-16
Update index.md

sheetalgulia:
I think this line is better. It is extremely important to use cache if your filter must retrieve data from the database, or other similar sources.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/555"></a>

### DIFF
--- a/docs/apis/plugintypes/filter/index.md
+++ b/docs/apis/plugintypes/filter/index.md
@@ -99,7 +99,7 @@ Filters are applied to all text that is printed with the output functions `forma
 
 ## Filter performance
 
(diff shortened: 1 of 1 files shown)


andrewnicols:
Hello,

Could you please explain why this is better? Gramatically it is correct as it is, and would be wrong with this proposed change.

sheetalgulia:
I think it is grammatically wrong and use caches would be better instead of uses caches.

andrewnicols (approved):

andrewnicols:
I've just re-re-re-re-read this and I agree. Thanks. Merigng.

PR #562, opened 2023-02-27
Update -8w task 5 for clarity

junpataleta:
* Sub-tasks reorganisation
* Since the "All must fix issues" filter now uses regex, there's no need to update it. We just need to make sure that the filters based on it look and work as expected.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/562"></a>


sarjona (approved):
Thanks Jun for improving this page. The changes look good :-)

PR #564, opened 2023-02-28
Update app docs

NoelDeMartin:


sarjona (approved):
Hi @NoelDeMartin!
Thanks for improving all these pages! It's great you've created separate commits for each change (it helps to process and understand them!). Well done!

PR #580, opened 2023-03-16
Update release process testing maintainer responsabilities

lameze:


general/development/process/release/index.md
```diff
@@ -19,13 +19,12 @@ sidebar_position: 7
 
 ## 7 weeks prior
 
-| # | Major | Minor | Task | Responsibility |
-| :---: | :---: | :---: | --- | --- |
+| # | Major | Minor | Task                                                                                                                                                                                                                                                                                                                                                                                                 | Responsibility |
```
sarjona, line 22: This change is not required (it was working properly before it was done). Is there any reason for adding all these lines?

sarjona (changes requested):
Thanks for working on this, Simey.

As you'll see, apart from the raised inline comment, there are a couple of things you should also review:

- Why you've added all these files in the [general/development/process/release/.idea/](https://github.com/moodle/devdocs/pull/580/commits/a2026c48982e3930f9571e2f3ec60c6f6a0102fe#diff-15a67033cd38b334fc11e4356309b85c74708ac7165af1c3110c80c851faeb3f) folder?
- As happened with the other issue, you should remove the unrelated commit (and leave only the one with the change) ;-)

data/libraries.json
andrewnicols, line 1: Can you please drop the unrelated changes from the PR.

general/development/process/release/index.md
```diff
@@ -23,9 +23,8 @@ sidebar_position: 7
 | :---: | :---: | :---: | --- | --- |
 | 1. | &#10003; |  | The [Continuous Integration](../integration/index.md#during-continuous-integration.2ffreeze.2fqa-period) period begins. Warn about it everywhere (telegram, exposed posts...).<br/>Rolling (on demand, beta, rc... all them together with stable weeklies) happens often (Tuesday & Friday are the usual days). All the team is on integration 100% since this week and until the end of continuous. | Integration Team |
 | 2. | &#10003; |  | Full demo of new code and sign-off for internal HQ projects. Decide which projects will be completed by the code freeze. | Platform manager / Head of educational solutions |
-| 3. | &#10003; |  | Check [closed qa_test_required-labelled issues](https://tracker.moodle.org/issues/?jql=labels%20%3D%20qa_test_required%20AND%20status%20%3D%20Closed) and create new QA tests as required. | Testing Maintainer |
```
andrewnicols, line 26: Doe this part of our process no longer exist then?

andrewnicols: Ah - I see... are these merged now? Do they need to be the same step? Isn't it just a case of changing from "Community manager" to "Testing Maintainer"?

lameze: I think having both in the same sections makes it easier to read

andrewnicols (approved):

andrewnicols (approved):

PR #590, opened 2023-03-29
docs: Add guidance on setting custom task data and user

andrewnicols:
When we wrote the task API we goofed by allowing just anyone set custom data. By setting custom data outside of the task consuming the data, we make the API harder to enforce:
* the parameters may change but we'd never find out without combing through logs
* formatting of the custom data is left to the caller

We should really encourage use of a factory to create the task itself.

I think that scheduling should still happen outside of the factory, but the user configuration also belongs in the factory.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/590"></a>

### DIFF
--- a/docs/apis/subsystems/task/adhoc.md
+++ b/docs/apis/subsystems/task/adhoc.md
@@ -51,26 +51,38 @@ class do_something extends \core\task\adhoc_task {
 
 </details>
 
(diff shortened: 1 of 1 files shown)


sarjona (approved):
Thanks for improving this documentation, Andrew! Well done!

PR #548, opened 2023-02-06
MDLSITE-6594 Add whitespace guidelines

junpataleta:
Adding whitespace guidelines, especially covering the agreed policy about whitespace before and after the assignment operator `=` in MDLSITE-6594.

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/548"></a>

PR #596, opened 2023-04-03
MDLSITE-7156 Back to original rules, but documented

stronk7:
This reverts partially #548 (MDLSITE-6594) because in MDLSITE-7156 it has been agreed to leave things like they were before both issues.

Added a final paragraph (common sense) about to take decisions always looking to context and similar code. I think it's acceptable.


junpataleta (approved):
Looks spot-on. Thanks, Eloy!

PR #625, opened 2023-05-16
Migrate the Profiling documentation

andrewnicols:
https://docs.moodle.org/dev/Profiling_PHP

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/625"></a>

### DIFF
--- a/data/migratedPages.yml
+++ b/data/migratedPages.yml
@@ -1494,6 +1494,9 @@ Privacy_API/Utilities:
 Process:
 - filePath: "/general/development/process.md"
   slug: "/general/development/process"
+Profiling_PHP:
+- filePath: "/docs/guides/profiling/index.md"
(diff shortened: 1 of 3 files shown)


timhunt:
Good to get this migrated. Are you planning to do a copy-edit. E.g. is it relevant to say PHP-7 these days?

Or do you want to save that for a future change?

andrewnicols:
Thanks Tim,

I thought I'd got them all but it was getting alte. I've updated the docs now to improve them generally, and have folded the https://docs.moodle.org/dev/Setting_up_xhprof_on_Moodle page into these docs too -- they only added very minimal additional information. I have elected not to copy any of those screenshots over just yet - they date back to the 'standard' theme and don't really add very much IMHO.

PR #634, opened 2023-05-23
MDL-77353 Functions for generating and displaying user details

tuanngocnguyen:
Update doc due to changes in

https://tracker.moodle.org/browse/MDL-77353

PR #635, opened 2023-05-23
MDL-77353 Functions for generating and displaying user details

tuanngocnguyen:


andrewnicols:
Hi @tuanngocnguyen ,

Is this the same as in #634?

sarjona:
Looking at #634 I realised both are about changes in MDL-77353, but this is changing docs/apis/core/user/index.md and #634 is changing docs/devupdate.md.

I've updated the patch here, to fix some of the issues raised by lint and it will be merged soon (once the jobs confirm it's OK). I've also backported it to versioned_docs/version-4.3/apis/core/user/index.md (because this issue was integrated in 4.3).

sarjona (approved):

PR #656, opened 2023-06-12
Add add_sticky_action_buttons to forms doc (MDL-77635)

ilyatregubov:


sarjona (changes requested):
Hola @ilyatregubov!
Thanks for updating this page. The text looks good but I have a question for you: shouldn't a reference to this method be added to docs/devupdate page? That way it will be easier for devs to be aware of its existence :-)

andrewnicols (approved):

andrewnicols (changes requested):
In addition to the devupdate, can we add a `<since>` tag to this please.

kthxbye :)

ilyatregubov:
Updated branch - added to  docs/devupdate page and added tag. I only added tag to forms page itself since  docs/devupdate page is for 4.3 on whole

sarjona (approved):
Thanks for working on this and adding the missing points (Since + dev update reference). It looks great so I'm happy to approve it :-)

sarjona (approved):

PR #693, opened 2023-07-26
MDL-78282 - [docs] Document sr_text output method

ferranrecio:
Document the new output methods created in MDL-78282

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/693"></a>

### DIFF
--- a/docs/apis/subsystems/output.md
+++ b/docs/apis/subsystems/output.md
@@ -313,6 +313,26 @@ Some interesting parameters for this function are:
 - `classes`: The classes of the paragraph. Note that this parameter is a comma-separated list of classes, not an array.
 - `id`: An optional id of the paragraph.
 
+#### sr_text()
+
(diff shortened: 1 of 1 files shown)


junpataleta (commented):
Thanks, Ferran. This looks good in general. A few points:
* It might be worth adding an example output of the function.
* Also, perhaps add a description about when it can be used or when it should not be used. E.g. the contents cannot be block elements as it will break HTML validation.

ferranrecio:
Hi @junpataleta ,

I amended the commit, adding an output example. However, I don't know in which cases it should not be used. Do we have any guide we can link to?

sarjona (approved):
Thanks for working on this, Ferran. The patch looks good and I think it's good enough to have it as it is because it can be improved in the future, if required, to add more information :-)

PR #699, opened 2023-08-11
MDL-78528 - [docs] Append a suffix to the completion rules

sarjona:
This patch adds information about the changes done in [MDL-78528](https://tracker.moodle.org/browse/MDL-78528).

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/699"></a>

### DIFF
--- a/docs/devupdate.md
+++ b/docs/devupdate.md
@@ -334,3 +334,49 @@ if (!empty($showonly)) {
     $mform->filter_shown_headers(explode(',', $showonly));
 }
 ```
+
+## Activity completion
+
+### Append a suffix to the completion rules
+
+As part of [MDL-78516](https://tracker.moodle.org/browse/MDL-78516), the [Default completion form](https://docs.moodle.org/en/Activity_completion_settings#Changing_activity_completion_settings_in_bulk) has undergone a significant rebuild to enhance code reusability and maintainability. To prevent duplicate IDs, a suffix has been introduced to the form elements related to completion rules.
+
+For third-party plugins, an adjustment is needed to incorporate this new suffix, following the approach already taken by [core modules](https://github.com/sarjona/moodle/commit/8f57f0fdaca027c7099bc6966467077aecbc0862).
+
+The primary modification entails editing `mod/yourplugin/mod_form.php` and applying the suffix to the completion rule elements within all relevant methods. As an example, here are the changes made to the `mod/choice` module:
+
+```php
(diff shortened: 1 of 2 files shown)


andrewnicols (approved):

docs/devupdate.md
```diff
@@ -334,3 +334,48 @@ if (!empty($showonly)) {
...
+
+As part of [MDL-78516](https://tracker.moodle.org/browse/MDL-78516), the [Default completion form](https://docs.moodle.org/en/Activity_completion_settings#Changing_activity_completion_settings_in_bulk) has undergone a significant rebuild to enhance code reusability and maintainability. To prevent duplicate IDs, a suffix has been introduced to the form elements related to completion rules.
+
+For third-party plugins, an adjustment is needed to incorporate this new suffix, following the approach already taken by [core modules](https://github.com/sarjona/moodle/commit/8f57f0fdaca027c7099bc6966467077aecbc0862).
+
+The primary modification entails editing `mod/yourplugin/mod_form.php` and applying the suffix to the completion rule elements within all relevant methods. As an example, here are the changes made to the `mod/choice` module:
+
+```php
+    public function add_completion_rules() {
+        $mform =& $this->_form;
```
andrewnicols, line 350: Can we remove the & from this. It's totally unnecessary :)

docs/devupdate.md
```diff
@@ -334,3 +334,48 @@ if (!empty($showonly)) {
...
+
+        $suffix = $this->get_suffix();
+        $completionsubmitel = 'completionsubmit' . $suffix;
```
andrewnicols, line 353: Can I suggest that we replace this duplicated string concate with a function:

```
$this->get_suffixed_name('completionsubmit');

protected function get_suffixed_name(string $fieldname): string {
    return $fieldname . $this->get_suffix();
}
```

The suffix concat is duplicated on half a dozen lines and we should aim to make our examples more readable.

docs/devupdate.md
sarjona: You're right!! Well spotted! :-)

docs/devupdate.md
sarjona: That's a great idea. I'll see if I can add this method to the trait in order to call it! ;-) If that's the case, I'll update again the docs in the future (but, for now, I've added it too).

andrewnicols (approved):

PR #706, opened 2023-08-21
[docs] Added 4.2.2 minor release security announcements

mickhawkins:


sarjona (approved):
Thanks for preparing the security announcements, @mickhawkins
The patch looks great! Well done!

PR #709, opened 2023-08-24
[docs] - plugin checklist approval blockers

danmarsden:
adds some common approval blockers to the checklist - in particular the security guidelines and significant namespace collisions are typical blockers.

@mudrd8mz good to have a +1 from you (or a -1) :-)

first dev docs PR so let me know if I've got something wrong...

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/709"></a>

### DIFF
--- a/general/community/plugincontribution/checklist.md
+++ b/general/community/plugincontribution/checklist.md
@@ -199,6 +199,8 @@ Examples of issues that will prevent your plugin from being approved:
 
 1. There is no public and transparent issue tracker where the community members can leave feedback, report bugs and suggest improvements.
 1. Your SQL fails to work on PostgreSQL even when working on MySQL.
+1. [Namespace collisions](../../development/policies/codingstyle/frankenstyle.md).
+1. Compliance with [security guidelines](../../development/policies/security/index.md).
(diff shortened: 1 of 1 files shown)


andrewnicols (approved):
This looks good to me. I'm happy to merge it now if you want? Or you can wait for David to take a peek first. Your choice? We can always land this and then create a new issue if there is more to add.

andrewnicols:
Ah - linting check is failing. This is something you can fix automatically:

```
yarn mdfix-all
```

Essentially this is complaining because you're referencing another document in the same section (general vs. docs). Where we do that we use a relative link to the markdown file because it allows better link checking.

mudrd8mz:
Many thanks @danmarsden for this. My only comment would be that the "Compliance with security guidelines" does not fit well into the "[something] that will prevent your plugin from being approved". The intention here was to list the approval blockers. So the blocker here could be something like "Violation of security guidelines" or so, no?

andrewnicols:
I've amended the branch and force-pushed it.

Note: Probably best not to branch it to `main`.

Issue #701, opened 2023-08-12
[docs] Create documentation for [Moodle 4.3 release]


stronk7:
### What is the Moodle feature that needs documenting?

With [Moodle 4.3 requirements (MDL-76724)](https://tracker.moodle.org/browse/MDL-76724) already applied upstream, it would be a good moment to create the https://moodledev.io/general/releases/4.3 page in advance and put there all the requirements, release notes... basically the 1st cut of the page.

Note that this is also related with the [PHP 8.2 Epic (MDL-76405)](https://tracker.moodle.org/browse/MDL-76405), because once we complete it, we'll have to edit both the Moodle 4.2.x and 4.3 pages.

I don't know when this is done normally, maybe later so, for your consideration.

### Is this documentation specific to a Moodle version?

Yes

### Are you able to provide a patch for this?

No (Default)

stronk7:
Closing as fixed by #718.

PR #718, opened 2023-08-29
Add 4.3.0 release information + metadata for minors due same day

andrewnicols:
Addresses #701


stronk7 (commented):
Hi @andrewnicols,

this looks perfect, only questions I've are:
1. Do we need to keep dragging the sodium, exif information to new versions?
2. Random thought, maybe we should "homogenise" the lang/words used in server requirements and database requirements and say something like "Not changed since Moodle X.Y", or "Changed in Moodle X.Y"... right now there is a bit of slight different lang/words used there.
3. It seems that the branch can be rebased (is outdated), the merge will do it, so no problem. Just telling if you want to do so.

andrewnicols:
> Do we need to keep dragging the sodium, exif information to new versions?

I'd say we don't need exif, because that's been recommended since before the 4.1 LTS.
For sodium I think we should keep it because it's a change since the last LTS (it was _recommended_ in 4.1).

> Random thought, maybe we should "homogenise" the lang/words used in server requirements and database requirements and say something like "Not changed since Moodle X.Y", or "Changed in Moodle X.Y"... right now there is a bit of slight different lang/words used there.

I thought I had already?

- For DB the language is all "increased since 4.1"
- For the 64-bit part, it's "changed since 4.1" because the word "increase" is not right IMO.

On the whole I think we should keep noting any changes which are different from before the last LTS, simply because the last LTS is a requirement for the _next_ upgrade. i.e. a lot of things need to change to upgrade from 4.0, to 4.1 (minimum php versions, DBs, etc.) but it's more _unusual_ to have some of those minimum version bumps between LTSs because of the nature of our rules.

> It seems that the branch can be rebased (is outdated), the merge will do it, so no problem. Just telling if you want to do so.

data/versions.json
```diff
@@ -1,6 +1,21 @@
 {
     "$schema": "../static/schema/versions.json",
     "versions": [
+        {
+            "name": "4.3",
+            "releaseDate": "9 October 2024",
```
stronk7, line 6: Aren't the years too much in the future? We are in 2023, if I'm not wrong, heh!

data/versions.json
```diff
@@ -1,6 +1,21 @@
...
+            "generalEndDate": "7 October 2025",
+            "securityEndDate": "5 May 2026",
```
stronk7, line 8: Shouldn't this be 2nd Monday of APRIL? (apart from the year).

general/releases.md
```diff
@@ -90,6 +90,17 @@ Minor releases dates differ slightly from release to release depending on the ti
 
 :::
 
+## Moodle 4.3
+
+<ReleaseTable releaseName="4.3" />
+
+:::info
+
+Bug fixes for general core bugs in 4.3.x will end 7 October 2025 (12 months).<br/>
```
stronk7, line 99: Again, the years.

general/releases.md
```diff
@@ -90,6 +90,17 @@ Minor releases dates differ slightly from release to release depending on the ti
...
+Bug fixes for security issues in 4.3.x will end 5 May 2026 (18 months).
```
stronk7, line 100: And again, the 2nd Monday of April (apart from the year).

stronk7 (changes requested):
All looks ok, just discovered a few small problems with the dates (years and days).

stronk7 (approved):

PR #721, opened 2023-08-31
MOBILE-4323 enrol: Document CoreEnrolDelegate for plugins

dpalou:


dpalou:
I opened this in the wrong repo, closing it now.

PR #733, opened 2023-09-13
Update app setup instructions

NoelDeMartin:


andrewnicols (changes requested):
Hi Noel,

Whilst these changes are correct, I wonder if it may be worth suggesting the use of Chrome for Testing:
https://developer.chrome.com/blog/chrome-for-testing/

It's a separate version of Chromium for the purpose of testing and with automatic updates disabled.

NoelDeMartin:
Thanks @andrewnicols for the suggestion, I wasn't familiar with "Chrome for Testing", I'll check it out.

However, for this PR, I think it's better if we go ahead with the merge because we're interested in getting it live before [the DevJam](https://moodlemoot.org/timetable/event/dev-jam/). We thought this 119 version of Chrome wouldn't be a problem until November, but we just realized it's already being distributed for Chrome canary and Chromium so it can potentially affect some people :(.

We could update it to suggest using Chrome for Testing. But considering that it's the first time we hear about it in our team and none of us has used it, I'm prefer to keep suggesting Chromium which is what we all use. Eventually, when we test it ourselves, we can update the documentation.

andrewnicols (approved):

PR #735, opened 2023-09-13
[docs] Update deprecation policy for named argument support

andrewnicols:
This is to solve issues identified and raised in MDLSITE-7358

<a href="https://gitpod.io/#https://github.com/moodle/devdocs/pull/735"></a>

### DIFF
--- a/general/development/policies/deprecation.md
+++ b/general/development/policies/deprecation.md
@@ -14,6 +14,8 @@ tags:
 
 ## Why is deprecation needed?
 
+import { ValidExample, InvalidExample, TabItem, Tabs } from '@site/src/components';
+
 In an open source project, the end use of the codebase varies. People may have customisations and plugins that depend on a function that has been targeted for deprecation. Rather than simply removing a function, we must gracefully deprecate the function over a period covered by a number of released versions.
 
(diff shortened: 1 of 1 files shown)


general/development/policies/deprecation.md
```diff
@@ -93,29 +95,49 @@ Longer deprecation periods can be considered for functions that are widely used.
...
+ * @param string[]|null $pets A list of pets that the individual has
+ */
+public function greet_person(
+    string $name,
+    null|int $age = null,
+    null|array $pets = [],
+}: void {
+    if ($age !== null) {
+        debugging(
+            'The age argument has been deprecated. Please remove it from your method calls.',
```
stronk7, line 123: “Parameter” vs “Argument” ?

stronk7 (dismissed):
Whops, sorry, clicked on start review instead of just comment.

We have been chatting about this… so… you know which are my only 2 suggestions (apart from the “argument” detail.

andrewnicols:
I think I've caught all uses of `arguments` instead of `parameters`, and I've tried to make it clear when the policy is effective from.

I've also provided an example of deprecating the method and replacing it.

junpataleta (approved):

PR #744, opened 2023-09-18
[docs] inline edit

crnfpp:


andrewnicols (approved):
I've added a couple of extra commits for the followign:

- we don't update the versioned docs pages with new content - they are their own system and don't have the pages you just added.
- update some broken links
- updating examples to have them render better:
 - narrow examples work better because they are easier to read without scrolling
 - whitespace between section is great because it improve legibility
- coding style fixes

PR #749, opened 2023-09-18
Moved general information to the top document

HoffmannP:


docs/guides/javascript/index.md
```diff
@@ -552,4 +552,52 @@ grunt
 </TabItem>
 </Tabs>
 
-### ESLint
+### JSHint
```
andrewnicols, line 555: We dont' actually use jshint any more. It is used by YUI but we don't useit, which is why it doesn't appear in this guide.

We will be removing YUI in the next 18-24 months (all being well) and will then only make use of eslint.

andrewnicols:
Closing as I believe this change is wrong.

PR #765, opened 2023-10-09
[docs] Update Moodle 4.2 references

sarjona:


mickhawkins:
Nice find @sarjona! Happy for this to go ahead if we just want to keep the current format, but I have a few suggestions that I think might make more sense (and reduce a bit of maintenance), since this relates to the page loading for "master" and not a specific version:

1. For the first 2 changes, change "Moodle 4.3" to "Moodle LMS".
2. The third change is commented out so is not visible on the page. I think we should either tweak it a bit so that it is just "You may want to read the Release Notes" (and link to the release notes homepage), or remove it entirely since it's not in use.

sarjona:
Hi @mickhawkins!
I think these changes were left with the Moodle version on purpose because that way, it's easy to identify the branch. If you look at the current versioned doc pages, you'll see what I mean:

- https://moodledev.io/docs/4.1
- https://moodledev.io/docs/4.2

Maybe this change should be done when the docs versioned folder was created but, as I wanted to ask tomorrow Andrew how to deal with it, I thought it was better to start preparing it and then tomorrow just duplicate the folder :-)

mickhawkins (approved):
I still think this is a generic page and shouldn't list any specific version, but given this is consistent with how it's been done in the past, I'll approve this so at least it's not listing an out of date version. We can always modify it later (eg when the 4.3 fork is done).

mickhawkins (approved):

mickhawkins (approved):

PR #768, opened 2023-10-10
Add testers of 4.3 QA cycle.

lameze:
As part of the QA cycle wrap up we need to update the list of testers credits


sarjona (approved):
HI @lameze!
Thanks for updating the QA data!! <3

PR #771, opened 2023-10-11
Writing acceptance tests: do not repeat feature-level tags for scenario

PhilippImhof:
Currently, the example repeats the `@javascript` tag in the scenario, even though it is already present in the feature's header. This is not necessary and would actually give a `duplicate-behat-tag` warning.

I suggest changing the example and adding a note that explains how tags defined at feature level will automatically apply to all scenarios of that feature.


stronk7 (approved):
Good detail, thanks.

PR #788, opened 2023-10-30
Peer review checklist improvements

timhunt:
Here are three proposed changes to the Peer review checklist, as three separate commits.

It might have been a mistake to send them as a single pull request, because while I think the first two should be uncontentious, the third one might need more discussion. If you want me to separate them out, just ask (or you can do it yourself).

The three changes are:

1. Move the overall checklist item to the end.
2. Rename it to 'Overall completeness and correctness' to more clearly name what it is for and because the Moodle content guideliness specifically [tell us to avoid the phrase 'sanity check'](https://moodledev.io/general/contentguidelines/styleguide/voice/contentprinciples#words-and-phrases-to-avoid).
3. Change the Accessibility check so that it is no longer 'optional'. That sent such a misleading signal about how much Moodle values accessibility, which is in reality a lot. (The reason it was 'options' is that people defined such a high bar for what was supposed to be a 'Peer review', so we felt it was unreasonable for it to be required. While accessiblity is importnat, and requires specialist knowledge, are other things like Security, Privacy and Performance. So, I think the Accessibility part of the peer review checklist should be brought into line - but I realise this might need discussion.

Obviously, if this gets merged, someone will need to update the Tracker comment template.

### DIFF
--- a/general/development/process/peer-review/index.md
+++ b/general/development/process/peer-review/index.md
@@ -235,34 +235,15 @@ Ensure that:
 
 ### Accessibility
 
-We aim for Moodle to be accessible to everyone, especially to people with disabilities. When peer-reviewing a patch that introduces changes on the frontend, it would be good to do a quick accessibility check on the page or on the UI elements affected by the patch.
+Moodle should be accessible to everyone. When reviewing any changes that affects the frontend, ensure that these points have been considered:
 
-Checking accessibility requires a bit of practice and knowledge, if you want to perform a systematic accessibility check, you can follow the [Accessibility checklist](./accessibility-checklist.md).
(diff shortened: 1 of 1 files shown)


junpataleta (commented):
Thanks for updating this, @timhunt. 

The changes look good to me, especially the updates for the accessibility section. I couldn't have worded those better. So thanks for that!

About removing `(Optional)` for the Accessibility check, I totally agree with you. +1000 for you to change that as well!

Once the issues from automated checks have also been addressed, I think this is good to go.

Cheers!

junpataleta (approved):
Looks like the automated issues have been fixed. Thanks, @andrewnicols !

timhunt:
Sorry, I missed one: we also need #790 mergecd.

PR #792, opened 2023-11-03
Add dev meeting for November 2023 and tidy up previous meetings

andrewnicols:


sarjona (changes requested):
The patch looks good and it's fantastic that you've reversed the order!
However, think the general/community/meetings page should be updated too, because it currently says: "Our next Developer meeting is October 2023 (date and time tbc)."

sarjona (approved):
Thanks Andrew! The patch looks great now!! :-)

Issue #779, opened 2023-10-19
Valid and invalid example components don't meet WCAG AA level


juliavs85:
### What is the issue that you're experiencing?

The validexample and invalidexample components fail the WCAG AA level because of the lack of colour contrast of the text.
See https://colourcontrast.cc/06a400/ffffff and  https://colourcontrast.cc/fa383e/ffffff
Changing the text to black fixes the issues. 

### Are you able to provide a patch for this?

None

PR #793, opened 2023-11-03
[repo] Meet WCAG AA level for Valid/Invalid example components

sarjona:
The validexample and invalidexample components fail the WCAG AA level because of the lack of colour contrast of the text. Changing the text to black fixes the issues.

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -16,6 +16,8 @@
   --ifm-color-primary-lighter: #d96900;
   --ifm-color-primary-lightest: #f67600;
   --ifm-color-contrasting: #194866;
+  --ifm-color-success: #008a00;
+  --ifm-color-danger: #ed060d;
   --ifm-font-family-base: "Open Sans";
   --ifm-navbar-height: 5.5rem;
   --ifm-navbar-padding-horizontal: 5.5rem;


andrewnicols:
Hi Sara,

Can you give an example of where we use these examples where it isn't _just_ code?

I'm seeing a contrast ratio of:
- 13.21:1 for the Invalid examples: https://webaim.org/resources/contrastchecker/?fcolor=4B1113&bcolor=FFEBEC
- 13.45:1 for the Valid examples: https://webaim.org/resources/contrastchecker/?fcolor=4B1113&bcolor=E6F6E6

andrewnicols:
These seem fine to me:
Screenshot 2023-12-12 at 9 52 36 am

sarjona:
Hi @andrewnicols!
Thanks for reviewing this. It was reported by Júlia and Jess: https://github.com/moodle/devdocs/issues/779
As you'll see, the title contrast (Do / Don't) is failing: https://colourcontrast.cc/06a400/ffffff and https://colourcontrast.cc/fa383e/ffffff

andrewnicols:
Any chance of some screenshots? I'm still not sure which bit we're referring to.

sarjona:
Sure!! For instance, the title Do / Don't in this page: https://moodledev.io/general/contentguidelines/productwriting/confirm-msg

Screenshot from 2023-12-12 07-49-39

andrewnicols:
This is the look with your current proposal:
Screenshot 2023-12-12 at 3 22 35 pm

This gets a -1 from me as it loses the aesthetic.

While the black on red technically passes WCAG AA contrast requirements, I personally find it fudge ugly and hard to read 😆 

Can I suggest something like this instead:
```
diff --git a/src/css/custom.css b/src/css/custom.css
index ce529e585..71db33918 100644
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -16,6 +16,8 @@
   --ifm-color-primary-lighter: #d96900;
   --ifm-color-primary-lightest: #f67600;
   --ifm-color-contrasting: #194866;
+  --ifm-color-success: #008a00;
+  --ifm-color-danger: #ed060d;
   --ifm-font-family-base: "Open Sans";
   --ifm-navbar-height: 5.5rem;
   --ifm-navbar-padding-horizontal: 5.5rem;
```
Screenshot 2023-12-12 at 3 24 31 pm

Might need to examine dark mode too.

sarjona:
Hi @andrewnicols!
I agree this is a better approach (and also accessible), so I've updated the patch accordingly :-)
With dark mode, it also looks good (and is accessible).

andrewnicols (approved):

PR #800, opened 2023-11-07
[docs] Fix plugins update page sidebar title and position

vmdef:
The Plugins directory updates page was recently created, but the Update link was not properly positioned in the sidebar and the title was not set. This patch sets the title and locates the Update link in the first position within the group.


sarjona (approved):
Thanks for improving the navigation for the Plugins directory, Víctor! :-)

PR #801, opened 2023-11-07
Update updates.md

federico-moodle:
The main title of the page is "Updates" H1 and then each update will be simply the date title H2


vmdef:
Hi @federico-moodle.

I have made this change in #800, therefore I believe we can close this one.

PR #808, opened 2023-11-13
[Docs] Updating an empty state example to make it clearer

juliavs85:


sarjona (approved):
The patch looks good Júlia! Well done :-) It's amazing to see your progress with git commands <3

PR #812, opened 2023-11-14
added recording for November meeting and updaged index page message

aurelie-moodle:


sarjona (approved):
Hi Aurélie! 
Thanks for updating these pages. I've fixed the errors and I'm going to integrate them :-)

PR #820, opened 2023-11-22
Use kinder language when talking about developers

cameron1729:


andrewnicols (approved):

abgreeve (approved):
I think that the change here looks reasonable. Discussion with developers through the public dev chat could be a way that a decision is made.

PR #822, opened 2023-11-27
Add unserialize() method to danger list.

paulholden:
Can lead to code execution exploits if not used properly with user supplied data. There are better methods of data exchange.

See also https://github.com/moodlehq/moodle-cs/pull/52


sarjona (approved):
Thanks for suggesting this improvement, @paulholden! Well done!

ferranrecio (approved):
Seems logic to me and it is consistent with our coding rules. I will merge now.

PR #830, opened 2023-12-05
added necessary step to composer installation

t-schroeder:
until there is a stable version of phpcompatibility that we can require

### DIFF
--- a/general/development/tools/phpcs.md
+++ b/general/development/tools/phpcs.md
@@ -22,6 +22,7 @@ We recommend use of the `moodle-extra` standard, particularly when writing new c
 The recommended method of installation is via the global composer command:
 
 ```console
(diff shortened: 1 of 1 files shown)


stronk7:
Thanks @t-schroeder ,

makes sense and matches what is [already documented @ moodle-cs](https://github.com/moodlehq/moodle-cs#using-composer-recommended), so LGTM.

stronk7 (approved):

stronk7:
OT, it's really curious that I've tried here installing it, both globally and in a temp dir... I've not got the "minimum-stability" set (neither globally, neither in the `composer.json` file itself).

There is some detail that I must be missing... anyway, let's hope a new release happens soon.

Ciao :-)

PR #837, opened 2023-12-13
[docs] Corrected 4.4 end security support date

mickhawkins:


sarjona (approved):
Thanks for updating this date, Mick! :-)

PR #838, opened 2023-12-14
Adding info on making a small change

wildgirl:


sarjona (approved):
@wildgirl , thanks for improving this part, to help people to make small changes in devdocs :-*

PR #850, opened 2024-01-04
[docs] Tidy up the devupdate to prioritise dev docs

andrewnicols:
This documentation was previously backwards and had the user-focused part of the documentation first.

The key part here is the _developer_ documentation regarding the method. The CSV format information is present to make testing easier and will also be present in the user docs.


ilyatregubov (approved):
The changes make sense for me. Thanks Andrew

PR #856, opened 2024-01-05
[docs] Improve wording for backport policy

andrewnicols:
I noticed an incorrectly terminated admonition, and then in fixing it tried to improve the language.

I also notethat there are only two files using "See also..." instead of "See also".

### DIFF
--- a/general/development/policies/backporting.md
+++ b/general/development/policies/backporting.md
@@ -37,7 +37,9 @@ The integration team will process backport requests, with the following guidelin
 If the backport request is approved, please follow the usual development process to submit the feature or improvement on earlier branches. Just to be clear, this means using the new bug number. So, even if the fix you are back-porting cherry-picks cleanly, you will need to amend the commit comment to use the new MDL-XXXXX number.
 
 :::tip
+
 It would be good practice to also add a line like "This is a backport of MDL-YYYYY." in the amended commit comment.
+
 :::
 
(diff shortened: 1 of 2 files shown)


general/development/policies/backporting.md
```diff
@@ -49,20 +51,19 @@ Many issues can be appropriately classified as borderline bug-fix/improvements.
 ## Backport fixes to unsupported branches
 
 - Given the [general policy](#general-policy) above, only supported stable branches are candidates normally.
-- Also security, privacy, data-loss and regressions caused by any of the previous issue types are accepted to be fixed into security-only supported branches.
-
-:::note
+- In addition to this any security, privacy, data-loss, and regression, caused by any of the previous issue types are accepted to be fixed into security-only supported branches.
```
stronk7, line 54: Is the comma after "data-loss" needed/correct? And, at the same time... shouldn't a comma be added after "previous issue types"?

Note this is how I'd say it in Spanish, heh. Not sure if the same rules apply. :-)

Other than that, my +1 to this. I think all the points in the policy remains correct.

andrewnicols: It's a separator in a list whose items are:
- security
- privacy
- data-loss
- regression

It's known as an Oxford comma. Personally I always use it, but others do not. It's neither right, nor wrong.

PR #857, opened 2024-01-07
Placeholder variable fixed in query example

yousuftafhim:


andrewnicols (approved):
Thanks  - good catch :)

PR #860, opened 2024-01-09
Fix missing description parameters in external_value constructor all

PhMemmel:


docs/apis/subsystems/external/writing-a-service.md
```diff
@@ -316,10 +316,10 @@ Because some web service protocols are strict about the number and types of argu
 ```php
 public static function get_biscuit_parameters() {
     return new external_function_parameters([
-        'chocolatechips' => new external_value(PARAM_BOOL, PARAM_REQUIRED),
-        'glutenfree' => new external_value(PARAM_BOOL, PARAM_DEFAULT, false),
+        'chocolatechips' => new external_value(PARAM_BOOL, 'if biscuit contains chocolate chips', PARAM_REQUIRED),
```
andrewnicols, line 319: If you don't mind, can you look at reformatting these as follows:

```suggestion
        'chocolatechips' => new external_value(
            PARAM_BOOL,
            'if biscuit contains chocolate chips',
            PARAM_REQUIRED,
        ),
```

The reason for this is that when it's a long line, it wraps and becomes less readable on Mobile.

Where we can, I try to wrap code samples at an earlier point for docs to make it more readable for those narrower screen situations.

docs/apis/subsystems/external/writing-a-service.md
```diff
@@ -316,10 +316,10 @@ Because some web service protocols are strict about the number and types of argu
...
+        'glutenfree' => new external_value(PARAM_BOOL, 'if biscuit is glutenfree', PARAM_DEFAULT, false),
```
andrewnicols, line 320: The other thing you can demonstrate is named args:

```suggestion
        'glutenfree' => new external_value(
            type: PARAM_BOOL,
            default: PARAM_DEFAULT,
            allownull: false,
        ),
```

andrewnicols: Also, this one is missing _two_ args, the sig is:

```php
    public function __construct(
        $type,
        $desc = '',
        $required = VALUE_REQUIRED,
        $default = null,
        $allownull = NULL_ALLOWED
    );
```

PARAM_DEFAULT is the 4th param, not the 3rd.

docs/apis/subsystems/external/writing-a-service.md
```diff
@@ -332,10 +332,10 @@ public static function get_biscuit_parameters() {
 public static function get_biscuit_parameters() {
     return new external_function_parameters([
         'ifeellike' => new external_single_structure([
-                'chocolatechips' => new external_value(PARAM_BOOL, VALUE_REQUIRED),
-                'glutenfree' => new external_value(PARAM_BOOL, PARAM_DEFAULT, false),
+                'chocolatechips' => new external_value(PARAM_BOOL, 'if biscuit contains chocolate chips', VALUE_REQUIRED),
```
andrewnicols, line 335: VALUE_REQUIRED also wrong.

andrewnicols (changes requested):
Apologies - they were worse than we all feared.

docs/apis/subsystems/external/writing-a-service.md
PhMemmel: why is this wrong?

docs/apis/subsystems/external/writing-a-service.md
PhMemmel: I used the named args, but your suggestion for `default: PARAM_DEFAULT` looks wrong to me. Shouldn't it be it `required: VALUE_DEFAULT, default: false`?

docs/apis/subsystems/external/writing-a-service.md
PhMemmel: sure, no problem

PhMemmel (commented):
Thanks for the review @andrewnicols!

I tried to include your suggestions.

Is it possible you were confusing `PARAM_` and `VALUE_` prefixes? Please have another look at my changes. From my point of view it now everything should be correct, but of course I can be wrong :)

docs/apis/subsystems/external/writing-a-service.md
andrewnicols: Here's the origial code:
```
        'chocolatechips' => new external_value(PARAM_BOOL, PARAM_REQUIRED),
```

And your version:
```
        'chocolatechips' => new external_value(PARAM_BOOL, 'if biscuit contains chocolate chips', PARAM_REQUIRED),
```

So there is no `PARAM_REQUIRED` - it's `VALUE_REQUIRED`

And for `glutenfree`:
```
        'glutenfree' => new external_value(PARAM_BOOL, 'if biscuit is glutenfree', PARAM_DEFAULT, false),
```

There is no `PARAM_DEFAULT`, it's `VALUE_DEFAULT`

So it's anyone's guess what the original code was trying to suggest. I suspect you're right, but that the contants are just incorrect and shoudl be VALUE_ not PARAM_.

docs/apis/subsystems/external/writing-a-service.md
andrewnicols: Apologies- I think this one is right. Just would benefit from switching to named args in php.

PhMemmel: so you are suggesting to switch to named args in the whole example? I just did it for 'glutenfree' to demonstrate that both is possible.

docs/apis/subsystems/external/writing-a-service.md
PhMemmel: Yes, true. As I said my whole changes now should fix this confusion in the whole example.

andrewnicols (approved):
That looks good to me, and demonstrates both options.

PhMemmel (commented):
Sorry, I forgot to push my commit, so they probably seemed to be pretty confusing :) Now pushed the commit.

docs/apis/subsystems/external/writing-a-service.md
```diff
@@ -316,10 +316,20 @@ Because some web service protocols are strict about the number and types of argu
 ```php
 public static function get_biscuit_parameters() {
     return new external_function_parameters([
-        'chocolatechips' => new external_value(PARAM_BOOL, PARAM_REQUIRED),
-        'glutenfree' => new external_value(PARAM_BOOL, PARAM_DEFAULT, false),
+        'chocolatechips' => new external_value(
+            PARAM_BOOL,
+            'if biscuit contains chocolate chips',
+            VALUE_REQUIRED,
+            VALUE_DEFAULT
```
andrewnicols, line 323: Ah - what is the `VALUE_DEFAULT` here? Fourth param is the _actual_ default value, so i think this was me getting everything confused and can come out.

As I say, why a move to named args is a really good thing going forwards.

PhMemmel: yes, overlooked that one. Removed the VALUE_DEFAULT.

PR #869, opened 2024-01-23
Add global definition of $CFG

PhMemmel:
Test class files do not seem to have access to global variables, so there is the need of declaring `global $CFG;` for this example to work.

Another possibility would be to use a relative path for the require_once statement. Not sure what is the better option in this case.


sarjona (approved):
Hi @PhMemmel!
Thanks for raising and fixing this! I've also applied it to 4.1, 4.2 and 4.3 versions of this file! :-)

PR #872, opened 2024-01-24
[docs] api: Fix string-deprecation admonition closing tag.

kabalin:
Fixing broken layout (https://moodledev.io/general/projects/api/string-deprecation#what-to-do-if-you-get-a-debugging-message):
image


kabalin:
Patched version of the layout: https://deploy-preview-872--moodledevdocs.netlify.app/general/projects/api/string-deprecation#what-to-do-if-you-get-a-debugging-message

sarjona (approved):
Thanks for raising and fixing this, @kabalin! Well done!

PR #875, opened 2024-01-26
[docs] Dev update BS5 reorganization

sarjona:
Added subsections into the BS5 section to help to easily find information.


roland04 (approved):
I like this new format @sarjona! Easier to read

PR #879, opened 2024-01-30
Update accessibility page

junpataleta:
* Add WCAG 2.1 Level AA accreditation date for 4.1/4.2
* Update the accessibility tools section to reflect the latest tools used for testing accessibility.


junpataleta:
Brett has reviewed the draft in Confluence, and I addressed the points he raised. I believe this is already in good shape and we can merge this. If we get further feedback, we'll just raise a new pull request.

jleyva (approved):

PR #880, opened 2024-01-30
[docs] Add intellectual property rights section

danmarsden:
Adding section of text requested by Chris Brown/

I think this is the right place for it, happy for suggestions to change.


junpataleta (approved):
Thanks, @danmarsden . This looks like the appropriate place to put this section.

PR #881, opened 2024-01-31
Group API: Fix typos in code examples

nero-92:
in the "How to get just the groups that the current user can see" example

### DIFF
--- a/docs/apis/subsystems/group/index.md
+++ b/docs/apis/subsystems/group/index.md
@@ -87,7 +87,7 @@ This is using an example from the module forums.
 $id = required_param('id', PARAM_INT);
 
 // Get the course module.
-$cm = get_coursemodule_from_id('forum', $id, 0, false, MUST_EXIST)
+$cm = get_coursemodule_from_id('forum', $id, 0, false, MUST_EXIST);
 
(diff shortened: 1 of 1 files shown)


junpataleta (changes requested):
Thanks for providing a patch, @nero-92. This looks great! 

Some comments:

1. It would be great to squash the commits into a single commit as if the patch was written perfectly for the first time.
2. It would be good to indicate the dev docs page that the patch affects. For example, something like `Group API: Fix typos in code examples` would be good.
3. I spotted another issue in the `How to find and use the "current" group` example code where there's a missing semicolon at the end of the [call to `get_coursemodule_from_id()`](https://github.com/moodle/devdocs/blob/main/docs/apis/subsystems/group/index.md?plain=1#L90). Might be worth fixing this in this pull request as well.

Hope this helps!

junpataleta (approved):
Thanks! This looks good now. We just need to squash the three commits into one and improve the commit message. I'll try to merge your pull request and see if the commits can be squashed into one.

junpataleta:
I took the liberty of squashing your commits. Thanks for your contribution!

PR #882, opened 2024-01-31
Add note to function deprecation about unit tests

marxjohnson:
Add a note to explain that asserDebugginCalled() should be added to unit tests that cover a deprecated function, to prevent them failing until the function is removed.


junpataleta (approved):
Good point. Thanks, @marxjohnson !

PR #883, opened 2024-01-31
Fix path to directory containing file converter plugins

geoffreyvanwyk:


junpataleta (approved):
Thanks, @geoffreyvanwyk . Looks spot-on!

PR #887, opened 2024-02-08
Add missing return statement

genius257:
Example code was missing required return statement.


junpataleta (approved):
Thanks, @genius257 ! Looks spot-on!

PR #891, opened 2024-02-19
[docs] Document decisions made in MDL-79077

andrewnicols:


docs/apis/core/hooks/index.md
```diff
@@ -18,6 +18,17 @@ The most common use case for hooks is to allow customisation of standard plugins
...
 
+## Hook Policies
+
+New hooks added to Moodle from Moodle 4.4 onwards must meet the following rules:
+
+- When describing a hook which happens before or after, the following modifiers are allowed:
+  - `before`
+  - `after`
+- When a modifier is used, it should be used as a _prefix_
+- Hooks belonging to a core subsystem should be located within that subsystem, and _not_ in the `core` subsystem
```
junpataleta, line 29: It would be good to improve this by providing several examples. 

For example
- Course-related hooks will be placed within `course/classes/hook`.
- Form-related hooks would be in `lib/form/classes/hook` given the form subsystem lives within `lib/form`
- Hooks from core APIs would be within `lib/classes/hook` (and probably recommended under their API folder for better organisation. For example, `core\navigation`-related hooks would be under `lib/classes/hook/navigation`)  

It would also be good if we could add examples that break the rules.

docs/apis/core/hooks/index.md
```diff
@@ -18,6 +18,17 @@ The most common use case for hooks is to allow customisation of standard plugins
...
+- Hook descriptions **should** be in English, and ** should not** use language strings or support translation
```
junpataleta, line 30: Super minor, extra space between `**` and `should not`

junpataleta (commented):
Thanks, @andrewnicols. Looks good in general. It would be great if we could have some valid and invalid examples for clarity.

andrewnicols:
I've added some examples and resolved that whitespace issue.

junpataleta (approved):

PR #902, opened 2024-02-23
[docs] Add docs for attribute_helper and attribute-described hooks

andrewnicols:
MDL-81011

### DIFF
--- a/docs/apis/core/hooks/index.md
+++ b/docs/apis/core/hooks/index.md
@@ -6,7 +6,7 @@ tags:
 - core
 ---
 
-import { Since, ValidExample } from '@site/src/components';
+import { Since, ValidExample, Tabs, TabItem } from '@site/src/components';
 
 <Since version="4.3" issueNumber="MDL-74954" />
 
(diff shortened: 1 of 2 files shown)


sarjona (changes requested):
Hi @andrewnicols!
Thanks a lot for working on this. The documentation is clear and I like the examples with both options (attributes vs described_hook interface). Well done!

However, looking at the patch in https://tracker.moodle.org/browse/MDL-81011, I miss having some references to the replaces_callbacks attribute you added later, after discussing it with Paul. Could you please review the docs to also include it? TIA!

andrewnicols:
Thanks Sara,

I've documented those now. Over to you!

docs/apis/core/hooks/index.md
```diff
@@ -341,23 +409,19 @@ To make use of Stoppable events, the hook simply needs to implement the `Psr\Eve
 
 namespace core\hook;
 
+use core\attribute;
+
+#[attribute\label('A hook dispatched just before a block instance is deleted')]
+#[attribute\hook\replaces_callbacks('pre_block_delete')]
```
sarjona, line 415: This example is confusing because is following both approaches, implementing interface and attribute. I was expecting a similar reference about deprecated_callback_replacement interface vs replaces_callbacks attribute, as you did for the described_hook vs label/tags attributes :-)

docs/apis/core/hooks/index.md
```diff
@@ -293,19 +368,12 @@ Since the hook is an arbitrary PHP object, it is possible to create any range of
 
 namespace core\hook;
 
-final class block_delete_pre implements described_hook, deprecated_callback_replacement {
-    public static function get_hook_description(): string {
-        return 'A hook dispatched just before a block instance is deleted';
-    }
-
+#[\core\attribute\label('A hook dispatched just before a block instance is deleted')]
+final class block_delete_pre implements deprecated_callback_replacement {
```
sarjona, line 372: Shouldn't an example for the replaces_callbacks interface should be added too?

sarjona (changes requested):
Hi @andrewnicols!
Thanks for working on these pages. devupdate.md looks great and, as you'll see, there are only a couple of minor things to improve in hooks/index.php :-)

PR #908, opened 2024-02-27
Remove task to remind devs about writing lang strings properly

junpataleta:
I have checked with Helen and we agreed that this task is not necessary and can be removed because wordings are already supposedly covered by the peer review stage:
https://moodledev.io/general/development/process/peer-review#language


sarjona (approved):
Hi @junpataleta!
Thanks for raising and improving this :-) The changes look good and make sense!

PR #912, opened 2024-03-07
[docs] Add notes about hook manager access via DI

andrewnicols:


sarjona (approved):
Hi @andrewnicols!
Thanks a lot for working on this. The information looks great, so I'm happy to approve it. Well done!

PR #920, opened 2024-03-18
[repo] Bump lighthouse timeout

andrewnicols:
Now that we have more supported branches in docs, it takes longer to build them.


sarjona (approved):
Hi @andrewnicols!
Thanks for fixing this (I've seen a few jobs that have failed recently because of this). Well spotted! <3

PR #922, opened 2024-03-18
[repo] Automatically assign author on PRs

andrewnicols:


stronk7 (approved):
Looks ok, and as far as it's working with other repos... makes sense.

PR #926, opened 2024-03-20
Add a missing transition to the CLR diagram

stronk7:
It's possible to send issues from "Waiting to Push" to "Waiting for CLR feedback".

This commit just adds that missing arrow (transition).

For reference, the diagram is available within the "MDL Workflow - Component Leads Review" folder in the HQ Integration collection (last page of the diagram).


ferranrecio (approved):
New image reviewed

PR #930, opened 2024-03-22
fix reference to the `deprecated` attribute

ziegenberg:
The deprecated attribute is located at `\core\attribute`.


sarjona (approved):
Thanks for reporting and fixing this. Well spotted, Daniel! <3

PR #935, opened 2024-03-26
Update 202403.md

OlgaKoenig1:
Adding the Meeting Recording


sarjona (approved):
Hi @OlgaKoenig1!
Thanks for adding the Meeting recording. I've changed the following:
- Updated the link with the one we can copy from the course (to redirect to the proper one depending on the user).
- Removed a couple of empty lines at the end of the file 
- Placed the Meeting recording in the "Details" section (for consistency with the rest of the meetings).
- Added a link on the Meetings page
- Updated the information about the next meeting (May, right?).

PR #937, opened 2024-03-26
[docs] Fix call to emit_deprecation_if_present using self

sarjona:


general/development/policies/deprecation/index.md
```diff
@@ -64,7 +64,7 @@ debugging('foobar() is deprecated. Please use foobar::blah() instead.', DEBUG_DE
 ```php
 #[\core\attribute\deprecated('foobar::blahv()', since: '4.4', mdl: 'MDL-XXXXX')]
 public function foobar(): void {
-    \core\deprecation::emit_deprecation_if_present([self, __FUNCTION__]);
+    \core\deprecation::emit_deprecation_if_present([self::class, __FUNCTION__]);
```
andrewnicols, line 67: Alternatively, use `$this`

```suggestion
    \core\deprecation::emit_deprecation_if_present([$this, __FUNCTION__]);
```

i think it's good to show a variety of options. Whilst `self` is wrong, we can show both `self::class` and `$this`.

andrewnicols (changes requested):

general/development/policies/deprecation/index.md
sarjona: I like the idea of using both! So, as you suggested, I've replaced this with `$this` <3

andrewnicols (approved):

PR #939, opened 2024-03-27
Update task api

brendanheywood:
Mostly little word tweaks, + a new section on admin setting checks


docs/apis/subsystems/check/index.md
```diff
@@ -225,6 +225,20 @@ https://github.com/moodle/moodle/blob/master/lib/classes/check/access/riskxss_re
 
 Some checks are by their nature asynchronous. For instance having moodle send an email to itself and then having it processed by the inbound mail handler to make it's properly configured (see [MDL-48800](https://tracker.moodle.org/browse/MDL-48800)). In cases like these please make sure the age or staleness of the check is shown in the summary, and you should also consider turning the result status into a warning if the result is too old. If appropriate make the threshold a configurable admin setting.
 
+## Reusing checks inside admin setting pages
+
+Quite often you have a plugin that includes some sort of status check in the admin settings page where you configure the plugin. Eg you can add the api keys for a service and it shows you right in the settings page if the connection settings work.
```
andrewnicols, line 230:
```suggestion
Quite often you have a plugin that includes some sort of status check in the admin settings page where you configure the plugin. For example you can add the api keys for a service and it shows you right in the settings page if the connection settings work.
```

We try to avoid `eg` and `ie` because they don't translate well to non-EFL speakers.

See:
- https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/e/eg
- https://developers.google.com/style/abbreviations#:~:text=Abbreviations%20not%20to%20use,-Prefer%20English%20terms&text=Don't%20use%20i.e.%20or,ymmv%2C%20RTFM%2C%20or%20others.

docs/apis/subsystems/check/index.md
```diff
@@ -225,6 +225,20 @@ https://github.com/moodle/moodle/blob/master/lib/classes/check/access/riskxss_re
...
+
+You can re-use the Check api to get this nice feature for free:
+
+```php
+$temp->add(new admin_setting_check('antivirus/checkantivirus', new \core\check\environment\antivirus()));
```
andrewnicols, line 235: I'd recommend wrapping this prematurely. I know the coding style is very liberal, but when viewed as code in docs it can lead to horizontal scrollbars being added.

You can see that here: https://deploy-preview-939--moodledevdocs.netlify.app/docs/apis/subsystems/check#reusing-checks-inside-admin-setting-pages if the window isn't obscenely large.

```suggestion
$temp->add(new admin_setting_check(
    'antivirus/checkantivirus',
    new \core\check\environment\antivirus(),
));
```

docs/apis/subsystems/check/index.md
```diff
@@ -225,6 +225,20 @@ https://github.com/moodle/moodle/blob/master/lib/classes/check/access/riskxss_re
...
+```
+
+These checks are performed asynchronously, and only if the check is visible, so will not slow down the whole admin tree which is often a downside of implementing this manually.
```
andrewnicols, line 238: Might be worth highlighting this with an admonition:

```suggestion
:::info

These checks are performed asynchronously, and only if the check is visible, so will not slow down the whole admin tree which is often a downside of implementing this manually.

:::
```

andrewnicols (changes requested):
Just a couple of minor tweaks to improve readability.

andrewnicols (approved):

andrewnicols:
Thanks Brendan,

I just rebased and force-pushed this to remove the weird Merge commit. Submitting for merge.

PR #940, opened 2024-03-27
[doc] tool_health was removed

ziegenberg:
See: https://tracker.moodle.org/browse/MDL-72335


sarjona (approved):
Hi @ziegenberg!
Thanks for fixing this. The health tool was removed from Moodle LMS core in Moodle 4.0, so I agree it's better to remove this section from the Check API subsystem. Well spotted! :-)

PR #942, opened 2024-03-28
Update index.md - Clarified documentation for string output functions.

michael-milette:


docs/apis/subsystems/output/index.md
```diff
@@ -198,66 +198,76 @@ For each of the functions below we'll try to explain when they should be used, e
 
 ### String formatting functions
 
+The format_string and format_text functions should always be used when preparing the output of information. They may also be used to process information before it is stored in the database however, filters should only be applied at output. For example, language filters must only be applied as the content is prepared for output because we don't yet know the user's preferred language.
```
andrewnicols, line 201:
```suggestion
The `format_string` and `format_text` functions should always be used when preparing the output of information. They may also be used to process information before it is stored in the database however, filters should only be applied at output. For example, language filters must only be applied as the content is prepared for output because we don't yet know the user's preferred language.
```

Generally best to put all method names in backticks.

docs/apis/subsystems/output/index.md
```diff
@@ -198,66 +198,76 @@ For each of the functions below we'll try to explain when they should be used, e
...
 
-These functions share the same code, so they will be explained together. The only difference is that `s()` returns the string, while `p()` prints it directly.
+The only difference between these two functions is that `s()` returns the string, while `p()` prints it directly.
 
 These functions should be used to:
 
 - print all the **values of form fields** like `<input>` or `<textarea>` tags.
 - **show plain (non-HTML) text** that has been introduced by the user (search string, quiz responses, ...).
-- print, in general, all the **dynamic data, not being HTML**, that doesn't need to be cleaned nor processed by filters. It is important not to use these functions for strings that contain HTML markup.
+- print, in general, all **dynamic data that is not HTML**, does not need to be cleaned or processed by filters. Never use these functions for strings that contain HTML markup.
```
andrewnicols, line 216: I wonder if we can rewrite this to be clearer. The double negative is not easy to parse.

```suggestion
- print, in general, all **dynamic HTML data must be cleaned or processed by filters**. Data that _can not_ contain HTML markup does not need to be processed. Never use these functions for strings that _might_ contain HTML markup.
```

It's also worth highlighting that it's anything that could possibyl contain HTML, even if that wasn't the intent. Anything whose PARAM_ type includes possible HTML.

docs/apis/subsystems/output/index.md
```diff
@@ -198,66 +198,76 @@ For each of the functions below we'll try to explain when they should be used, e
...
 
 The key parameter for this function is:
 
-- **strip**: it decides if we want to strip slashes from the string or not. By default, it's `false`, so no strip will be performed. We should set such parameter to 'true' only when data to be processed isn't coming from database, but from HTTP requests (forms, links, ...).
+- **strip**: Set to `true` to strip slashes from the string. Only set this parameter to `true` when the data to be processed is not coming from the database. This should be used when the string comes from HTTP requests (forms, links, ...). (Default is `false`, so no strip will be performed)
 
 #### format_text()
 
+```php
+function format_text(string $text, $format = FORMAT_MOODLE, ['noclean' => false, 'trusted' => false, 'filter' => true, 'context' => $context, 'param' => true, 'newline' => true, 'allowid' => false, 'blanktarget' => false])
```
andrewnicols, line 227: I really don't think we should include funciton signatures in here. They change.

If we _do_, then we should format them in a readable way, and we should encourage the use of named parameters where possible.

```suggestion
function format_text(
    string $text,
    $format = FORMAT_MOODLE,
    [
        'noclean' => false,
        'trusted' => false,
        'filter' => true,
        'context' => $context,
        'param' => true,
        'newline' => true,
        'allowid' => false,
        'blanktarget' => false,
    ],
);

```

andrewnicols: We could also start to recommend that people consider using `\core\formatting` directly instead, e.g.

```php
\core\di::get(\core\formatting::class)->format_text(
    ?string $text,
    string $format = FORMAT_MOODLE,
    ?context $context = null,
    bool $trusted = false,
    ?bool $clean = null,
    bool $filter = true,
    bool $para = true,
    bool $newlines = true,
    bool $overflowdiv = false,
    bool $blanktarget = false,
    bool $allowid = false,
): string;
```

docs/apis/subsystems/output/index.md
```diff
@@ -198,66 +198,76 @@ For each of the functions below we'll try to explain when they should be used, e
...
+  - `options->filter`: Set to `false` if you do not want to allow filters to process the text. This is ignored by `FORMAT_PLAIN` for which filters are never applied. (Default is `true`)
+  - `options->context`: If text is filtered (and this happens by default), it is very important to specify the context (id or object) for applying filters. If context is not specified it will be taken from `$PAGE->context` and may potentially result in displaying the same text differently on different pages. For example, all module-related information should have module context even when it appears in course-level reports, all course-related information such as name and description should have course context even when displayed on the front page or system pages.
+  - `options->param`: Set to `false` if you do not want every paragraph to automatically be enclosed between HTML paragraph tags (`<p>...</p>`). This option only applies to `FORMAT_MOODLE`. (Defaults is `true`)
```
andrewnicols, line 244: This one is actually `para`, not `param` - as in paragraph

andrewnicols (changes requested):

andrewnicols (approved):

PR #945, opened 2024-04-05
Fix some errors on hooks api page

PhMemmel:
While implementing my first hook I stumbled over these - IMO - errors on the hook API page.

### DIFF
--- a/docs/apis/core/hooks/index.md
+++ b/docs/apis/core/hooks/index.md
@@ -158,9 +158,9 @@ namespace mod_activity\hook;
 
 #[\core\attribute\label('Hook dispatched at the very end of installation of mod_activity plugin.')]
 #[\core\attribute\tags('installation')]
-class installation_finished implements \core\hook\described_hook {
+final class installation_finished {
     public function __construct(
-        public function string $version,
+        public readonly string $version,
     ) {
     }
 }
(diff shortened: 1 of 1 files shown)


docs/apis/core/hooks/index.md
```diff
@@ -158,9 +158,9 @@ namespace mod_activity\hook;
 
 #[\core\attribute\label('Hook dispatched at the very end of installation of mod_activity plugin.')]
 #[\core\attribute\tags('installation')]
-class installation_finished implements \core\hook\described_hook {
+final class installation_finished {
     public function __construct(
-        public function string $version,
+        public string $version,
```
andrewnicols, line 163: Bah - this was mean to be using readonly properties, but I mistyped `function` instead.
```suggestion
        public readonly string $version,
```

docs/apis/core/hooks/index.md
```diff
@@ -174,9 +174,9 @@ class installation_finished implements \core\hook\described_hook {
 <?php
 namespace mod_activity\hook;
 
-class installation_finished implements \core\hook\described_hook {
+final class installation_finished implements \core\hook\described_hook {
     public function __construct(
-        public function string $version,
+        public string $version,
```
andrewnicols, line 179: Same here

andrewnicols (changes requested):
Thanks @PhMemmel ,

Good catch on those. The `public function` were intended to be `public readonly` and be demonstrating constructor property promotion. D'oh!

PhMemmel:
Requested changes added

andrewnicols (approved):

PR #958, opened 2024-04-09
Only reference current versions

fulldecent:
This removes references to old versions, which should not be wanting to develop for, because they are unsupported


andrewnicols (approved):
Note: These are only examples - the actual versions do not matter.
We do not really need to keep these up-to-date.

PR #960, opened 2024-04-11
Update releases.md to be less clear about what is supported.

fulldecent:
Support: https://tracker.moodle.org/browse/MDLSITE-7603?focusedId=1075641&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1075641

### DIFF
--- a/general/releases.md
+++ b/general/releases.md
@@ -69,6 +69,8 @@ The most recent [long-term support release (LTS)](https://en.wikipedia.org/wiki/
 </div>
 </details>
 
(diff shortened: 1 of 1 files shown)


andrewnicols:
This isn't really worht mentionining.

We have regression versions in the tracker so that we know which version that it was reported against, and we keep the previous version (currently 3.9 and 3.11) for a very short period (a matter of weeks if I recall) after their final release to allow for recently introduced serious bugs to be reported and a possible unplanned release to be made against those versions, but those cases are so rare that it simply isn't worth mentioning.

PR #961, opened 2024-04-11
[repo] Adjust /docs to include next version

### LINKED ISSUE #955: [chore] Change default path of next release for Moodle 4.5 onwards
### What do we need to do?

So as we're going to be releasing Moodle 4.4 soon and we'll start development on 4.5, I am wondering whether we should change the default path for the in-development version of the docs.

Right now we have versioned paths for released Moodle versions, that is:
- /docs/4.3/apis
- /docs/4.2/apis
- /docs/4.1/apis

And we have the 'current' version (in development - main branch). That is at:
- /docs/apis

This works fairly well but it does ahve a main pitfall, being that any link to 4.4 docs may disappear in the future as that file may be moved/etc.

The prime example is the `devupdate.md` file. That gets reset with each release and so the bookmark within it will (likely) change.

The alternative is that we set the path for the 'current' version to the next version of Moodle. So since Moodle 4.4 is our current development branch, we'll use that example:
- /docs/4.3/apis - as before. 4.3 API docs
- /docs/4.2/apis - as before. 4.2 API docs
- /docs/4.1/apis - as before. 4.1 API docs
- /docs/4.4/apis - the docs for the _next_ version of Moodle.

The drawback here is that there is no `/docs/apis` any more. I believe that we can add a redirect so that any page not found will redirect to the 4.4 path.

andrewnicols:
This commit changes the unversioned docs at /docs/ to instead be at /docs/[nextVersionNumber].

It includes a redirect for /docs/* (where an existing page is not found) and it adds helpers to specify the version number in URLs and other locations.

Fixes #955

### DIFF
--- a/.markdownlint/fix-renamed-links.js
+++ b/.markdownlint/fix-renamed-links.js
@@ -123,6 +123,14 @@ const isRelativeLinkPossible = (link, file) => {
     const linkFolder = link.split('/')[1];
     const fileFolder = file.split('/')[1];
 
+    if (linkFolder === 'docs' && fileFolder === 'docs') {
+        if (link.match(/^\/docs\/\d\.\d\//)) {
+            // This is a versioned link.
+            // We don't want to normalise these.
+            return false;
+        }
+    }
+
     // In Docusaurus, different root folders contain different plugin instances.
     // Docusaurus does not support relative links between different plugins.
     return linkFolder === fileFolder;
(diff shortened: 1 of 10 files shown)


andrewnicols:
@stronk7 can you check if this behaves how you had in mind?

stronk7:
So, keeping already versioned docs (<= 4.3 right now) that will continue working EXACTLY the same, what we are doing here is:

1. The current docs place in the repo continue being exactly the same (unversioned).
2. We stop serving them in "unversioned" URL and always redirect for the to "versioned" URL (no matter it still doesn't exist as versioned doc). So `https://moodledev.io/docs/xxx` is redirected to `https://moodledev.io/docs/4.4/xxx` (no matter it's still served from unversioned docs.

And that's the unique change, the day that the 4.4-versioned copy is done, immediately then, `https://moodledev.io/docs/xxx` will be redirected to the next version `https://moodledev.io/docs/4.5/xxx`. And versioned will work for <= 4.4.

It sounds good for me. So, for links that we want to be auto-updated, we always use the unversioned URL and that will, automatically, redirect to the current one under development (/4.4/ right now, /4.5/ in a few weeks, ...). And whenever we want to point to a versioned one, we just add the version to the URL.

Ciao :-)

sarjona (approved):
Hi @andrewnicols!
Thanks a lot for working on this improvement!!! It's great that, apart from implementing the changes for using the next docs version by default, you've also improved the performance! Well done!! <3

I'm happy to approve and merge these changes! :-)

PR #963, opened 2024-04-11
[repo] Add docusaurus cache

andrewnicols:


kabalin:
Wondering, if we need to specify `BUILD_FAST` env var, which seems reduce test build to current version of docs only, Docusaurus is using it in [GH workflows](https://github.com/search?q=repo%3Afacebook%2Fdocusaurus%20build%3Awebsite%3Afast&type=code).

andrewnicols:
I'd like to avoid the BUILD_FAST as long as possible just because it means we can't review changes on backported docs.

sarjona (approved):
Thanks for adding this improvement, Andrew! And also for clarifying the question that Ruslan raised ;-)
The patch looks good and I'm happy to merge it :-)

PR #965, opened 2024-04-12
Add example/tutorial repo mod_nosferatu for reactive library page

Menrath:
This repository turned out to be incredibly helpful for me, I really think it should be referenced on top of the developer resources page for the new reactive library.


sarjona (approved):
Hi @Menrath!
Thanks a lot for adding this example to the reactive library page! Good idea! 
I fixed the problem with the empty line and replaced the new section with a tip admonition.

I'm happy to approve and merge it! :-)

PR #968, opened 2024-04-16
Bump the 1 week prior point 13 to actual MRs

stronk7:
Also, apply a little fix because we enable the lang tests when the NEXT (aka, 4.5) lang packs are available, not when the current release (4.4) ones are available. The later are available since 6 months ago.


sarjona (approved):
Hi @stronk7!
Thanks for reviewing, updating and improving the release process! 
The changes look good so I'm happy to approve and merge them. Well done!

PR #978, opened 2024-04-20
Update issues to clone for the next release

junpataleta:


andrewnicols (approved):
Suggest holding this til Monday at 9am to trigger a final build of docs with new versions

PR #980, opened 2024-04-22
4.5 dates

andrewnicols:


andrewnicols:
Dates pulled from the product heartbeat. I think I have the correct dates for end of support too.

src/components/ReleaseTable/index.tsx
sarjona, line 1: Hi @andrewnicols!
Thanks for working on this. The changes look good but I'm not sure about this one because, although I find it useful having when the next minor will be released for each Moodle version... Is there any other reason to hide this information from users? Otherwise, I'll suggest leaving it as it is now (because I know it helps people to find the dates).

sarjona (changes requested):

sarjona (approved):
Hi @andrewnicols!
Thanks for removing the commit to hide the next releases :-)
The patch looks great and it's working as expected ;-)

PR #990, opened 2024-04-23
Remove tideways references from Profiling documentation

marxjohnson:
The tideways_xhprof extentension is no longer maintained, from the GitHub README:

> This XHProf PHP extension fork has outlived its purpose and is 
> archived in favor of the revitalized PECL XHProf extension
> that is also using the modern PHP profiling abilities and 
> performant timer APIs.


sarjona (approved):
Hi @marxjohnson!
Thanks for reporting and improving this :-) The patch looks good and makes sense.

I've also created https://tracker.moodle.org/browse/MDL-81646 to remove references from Moodle LMS ;-)

PR #1015, opened 2024-05-08
[docs] Update 2024/05 agenda with more detail

andrewnicols:


timhunt (approved):
Looks good to me. Ready to merge.

Andrew, you could consider also (re-)covering some 4.4 changes. Specifically dependency-injection. I know you covered it once before, but for the audience "heard about it once" and "know what to do" are two different things. So, if you have time, you might consider adding a re-visit of that, particularly if there is emerging good practice which everyone should be following. Thanks for considering this.

PR #1018, opened 2024-05-10
Add useful information about code freeze and other key dates.

### LINKED ISSUE #1017: Make ie easier to find the freeze dates for the next release or two
### What is the Moodle feature that needs documenting?

It may just be me, but I always struggle to find the freeze dates for the next Moodle release. I know they are in the Moodle Development Calendar, but that is not easily browsable. At least, the link on https://moodledev.io/general/releases seems to be an iCal download.

Andrew asked "I’ll look at adding that metadata to our versions json and putting it into the releases page or elsewhere. [Please] Can you create a devdocs issue?" so here it is.

### Is this documentation specific to a Moodle version?

None

None

andrewnicols:
Fixes #1017 

Note: I think we should add additional data here, including:
- next minor date
- supported PHP versions
- link to issue deciding requirements

Those are for the future.

### DIFF
--- a/.markdownlint-cli2.cjs
+++ b/.markdownlint-cli2.cjs
@@ -176,6 +176,7 @@ config.MD033 = {
         "DbAccessPHP",
         "SettingsPHP",
         "LocalLib",
+        "ReleaseStateSummary",
     ]
 };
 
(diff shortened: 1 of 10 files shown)


sarjona (approved):
Hola @andrewnicols!
Thanks a lot for working on this improvement. The patch looks good and the information displayed is super-useful! Well done!
Kind regards!

PR #1019, opened 2024-05-12
Added missing comma

Bodoque-dios:


andrewnicols:
Hi @Bodoque-dios ,

This change looks correct. Are you able to apply it to the other affected versions (in `versioned_docs/*`) and the 4.5 docs (`/docs/*`)? If you can squash the commits into a single commit when you do so that would be ideal.

Thanks,

Andrew

sarjona (approved):
Hi @Bodoque-dios!
Thanks for fixing this and considering Andrew's comments :-) The patch looks good so I'm happy to approve it ;-)

PR #1021, opened 2024-05-16
Update index.md

TobiGa:
Add missing class declaration.


sarjona (changes requested):
Hi @TobiGa!

Thanks for fixing this. The change looks correct. Are you able to apply it to the other affected versions (in versioned_docs/) and the 4.5 docs (/docs/)? If you can squash the commits into a single commit when you do so that would be ideal.

Cheers!
Sara

TobiGa:
Hi @sarjona,
sure! Added the change to all the places i could find and squashed the result.

Cheers
Tobi

andrewnicols (approved):

PR #1024, opened 2024-05-17
[docs] Document use of legacyclasses.php

andrewnicols:
MDL-81919

### DIFF
--- a/docs/devupdate.md
+++ b/docs/devupdate.md
@@ -47,6 +47,27 @@ Please note that the same limitations regarding access to the Database, Session,
 
 :::
 
+#### Autoloading legacy classes
+
+<Since version="4.5" issueNumber="MDL-81919" />
+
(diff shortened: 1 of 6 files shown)


docs/devupdate.md
```diff
@@ -47,6 +47,27 @@ Please note that the same limitations regarding access to the Database, Session,
 
 :::
 
+#### Autoloading legacy classes
+
+<Since version="4.4" issueNumber="MDL-81919" />
```
junpataleta, line 52: Should be since 4.5

junpataleta (changes requested):
Thanks, @andrewnicols. Looks pretty much spot-on. We just need to set the correct version to `Since 4.5` instead of `Since 4.4`. Feel free to fix up the `ABORT_AFTER_CONFIG` one too.

Once this is sorted, I think this is good to go. Cheers!

junpataleta (approved):

PR #1029, opened 2024-05-22
[docs] Add documentation for new SMS API

andrewnicols:


mickhawkins:
Hi @andrewnicols,

This is great! Just a couple of minor suggestions:

1. Under fetching messages, it would be useful to mention the message body will not be stored if the issensitive flag is true, and that those messages can only be sent synchronously for that reason.
2. Should the example for get_messages (plural) use a named argument?

mickhawkins (approved):

PR #1033, opened 2024-05-30
Database activity sticky footer dev update (4.5).

paulholden:
https://tracker.moodle.org/browse/MDL-81321


andrewnicols (approved):

sarjona:
This is documented in the patch and will be included in UPGRADING.md. As the iTeam agreed, it does not need to be added to devupgrade.md as well so I'm going to close this issue.

PR #1063, opened 2024-06-27
[docs] Improve generator tool using behat - MDL-81885

ferranrecio:
MDL-81885 allows admin_generator to execute the "the_following_config_values_are_set_as_admin behat step". Prior to this patch, the tool could only execute the standard generator steps.

I also moved the execution process to its own section so it can be linked directly to the issues when needed.


sarjona (approved):
Thanks for working on this Ferran. It's great that, apart from adding a reference to the new config settings supported once MDL-81885 is integrated, you also improved the page so, from now on, we can link it from testing instructions each time we provide a feature file to initialise testing instructions. Well done! :-)

I'll merge this pull-request once weekly is rolled :-)

PR #1066, opened 2024-06-28
Updated view-php.tsx

hamidayubdev:
`get_course_and_cm_from_cmid` can run into exception if `cmid` doesn't exist, thus added try catch block in view-php.tsx.


andrewnicols:
This isn't somethign that we typically do. That's not to say that people _can't_ but it's not something done in most situatinons. Additionally the `print_error()` method is deprecated and shouldn't be used any more.

PR #1067, opened 2024-06-29
Refactored Index.js of theme/navbar

hamidayubdev:
Simplified logic of returning Logo component in `theme/navbar/logo/index.js` by removing empty early return.

Refactored openMenu function in `theme/navbar/index.js`.


andrewnicols (commented):
Sorry but this doesn’t work in all situations.

Our version numbers have three parts and this normalises them to extract just the first two.

hamidayubdev:
@andrewnicols Which one is not working? 
This (refactored getVersion function to directly use versionName ...) commit?

PR #1071, opened 2024-07-01
Add information about CSS styles to Tiny plugin doc

sammarshallou:


andrewnicols (changes requested):
It is probably worth noting that hsould be really f'king careful about how much style you add.

You do _NOT_ want to have styling for a specific editor plugin in _content_. If that plugin is ever removed, it will break the content.

Worth adding an admontion:

```markdown
:::note

Care should be taken when adding content styles so that generated content is not tied to the Editor Plugin. If the plugin is removed, it should not break content.

:::
```

sammarshallou:
thanks Andrew, I slightly updated your suggested text because the problem would also occur if individual users switch editor (eg to Atto which doesn't use the editor_styles.css)

andrewnicols (approved):

andrewnicols:
LGTM - thanks! :)

PR #1074, opened 2024-07-04
[docs] Add reference to public files deprecation

sarjona:
As we've discussed with each deprecation of a public file, like in MDL-43938 with badges/newbadge.php, I've added an explicit reference to it. This will help us remember the agreed-upon process in the future.


aanabit (approved):
I agree with the proposed deprecation approach and doc changes.

PR #1084, opened 2024-07-19
Import ModalRegistry instead of ModalFactory

omernawaz:
We need to import ModalRegistry not ModalFactory when defining a new Modal Class for creating custom modal types


andrewnicols (approved):
Thanks for catching this. I've taken the liberty of applying your patch to 4.1 too.

PR #1091, opened 2024-08-01
Link to the statement of conformance file

junpataleta:
WebKey IT used to list its clients' accreditations on its website. However, since it was acquired by AbleDocs and later became GrackleDocs, it no longer seems to update this list.

Because of this, the link to the "WCAG 2.1 Level AA conformance" in our dev docs points to an old accreditation. I think it's better to replace the link with a URL to the actual statement of conformance that is currently on the 4.2 accessibility audit epic (MDL-78185).

### DIFF
--- a/general/development/policies/accessibility.md
+++ b/general/development/policies/accessibility.md
@@ -16,7 +16,7 @@ As part of our ongoing commitment to accessibility and continuously improving co
 
 #### Moodle LMS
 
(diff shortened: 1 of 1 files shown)


sarjona (approved):
Hi @junpataleta!
Thanks for updating this URL. I agree that it's better and the only problem I can see linking this PDF file is that we need to remember to update it every time the accreditation is renewed (but I'm sure you'll keep that in mind).

I'm happy to approve and merge it!

PR #1100, opened 2024-08-13
Specify value for XDEBUG_PROFILE in profiling page 

ctam:
Specify the value for `XDEBUG_PROFILE` param to solve this issue: The trigger value '', as set through 'XDEBUG_PROFILE', did not match the shared secret (xdebug.trigger_value) for mode 'profile'.

### DIFF
--- a/docs/guides/profiling/index.md
+++ b/docs/guides/profiling/index.md
@@ -148,9 +148,9 @@ xdebug.trigger_value = ProfileMe
 
 :::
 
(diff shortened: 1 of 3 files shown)


sarjona (approved):
Hi @ctam!
Thanks for helping us to improve this page. I also changed versioned_docs/version-4.3/guides/profiling/index.md and docs/guides/profiling/index.md to guarantee all the versions for this page are updated properly :-)

PR #1104, opened 2024-08-21
Document named_templates

NeillM:
We probably want to let people know about named templates and the ability to save on writing some boiler plate code when using templates in the Output API


NeillM:
I realised I needed to update more places.

andrewnicols (approved):

PR #1108, opened 2024-08-28
Clarify the requirements for backporting

junpataleta:


general/development/policies/backporting.md
```diff
@@ -41,7 +41,7 @@ Should you feel that a new feature or improvement needs backporting, please foll
 The integration team will process backport requests, with the following guidelines:
 
 1. The integration team will together consider each request individually considering the needs of the community (influenced by forum posts, moodle partners, discussion with developers via tracker or private message, etc).
-2. Backports will happen not earlier than 3 weeks and not later than 2 months after the request was performed.
+2. Backport requests can be considered not earlier than 3 weeks and not later than 2 months after the original improvement/new feature has been integrated.
```
snake, line 44: Just pointing out that, perhaps, it's permitted that they're consider by iteam before the 3 weeks (say if one is submitted the same day as the feature is integrated or inside the first 3 weeks), but we won't action (i.e. the backport won't happen) it until such time as +3 weeks.

sarjona (approved):
Hi @junpataleta!
Thanks for raising and improving this. I think the proposal is clearer so I'm happy to approve it :-)

PR #1112, opened 2024-09-02
[docs] Added new parameter @notify to file creation function - MDL-75850

meirzamoodle:
Note: This is blocked by [MDL-75850](https://tracker.moodle.org/browse/MDL-75850)

### DIFF
--- a/docs/apis/subsystems/files/index.md
+++ b/docs/apis/subsystems/files/index.md
@@ -293,6 +293,19 @@ $fs = get_file_storage();
 $fs->create_file_from_storedfile($fileinfo, $existingfile);
 ```
 
+#### Notification requirements for new files
+
+By default, the `$notify` parameter has a value of `true` and will call the after_file_created hook after files are created. If you prefer not to do this, you can set it to `false`.
+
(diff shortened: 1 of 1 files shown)


docs/apis/subsystems/files/index.md
```diff
@@ -293,6 +293,19 @@ $fs = get_file_storage();
 $fs->create_file_from_storedfile($fileinfo, $existingfile);
 ```
 
+#### Notification requirements for new files
+
+By default, the `$notify` parameter has a value of `true` and will call the after_file_created hook after files are created. If you prefer not to do this, you can set it to `false`.
```
ilyatregubov, line 298: This is very unclear. What parameter? Of which function? What are we trying to achieve? I would mentioned the function directly and explained a new argument meaning - thery is no connection between argument name notufy and the fact that we are calling come hook. And same appies to the section name - what requirement is meant?

ilyatregubov (changes requested):

meirzamoodle:
I am closing this one because there are changes that make the notification unnecessary.

PR #1113, opened 2024-09-03
[docs] document some activity [modname]_supports features

ferranrecio:
Since MDL-81767, there is a new FEATURE_QUICKCREATE that needs to be documented.


sarjona (approved):
Hi @ferranrecio!
Thanks for documenting the new FEATURE_QUICKCREATE and also for improving the _supports() method. Well done!! It looks great!! <3

PR #1120, opened 2024-09-12
[general] Add September 2024 meeting

sarjona:
Pending update date and agenda (@wildgirl will do) ;-)


sarjona:
I'm merging this page because Helen will update the content :-)

PR #1122, opened 2024-09-12
Moving September meeting link to top, as it's not yet a past meeting

wildgirl:


sarjona (approved):
Hi Helen!
Well spotted! The patch looks good (it was failing because there was a trailing space that I've removed) :-)

PR #1124, opened 2024-09-13
Improve the 'testing' part of the peer review checklist

timhunt:
This edit is in response to having to keep saying the same thing about changes that have come to me to CLR, after someone claims to have peer reviewed them. I hope this is not too opinionated.

### DIFF
--- a/general/development/process/peer-review/index.md
+++ b/general/development/process/peer-review/index.md
@@ -113,7 +113,7 @@ Ensure that:
 
 ### Testing instructions and automated tests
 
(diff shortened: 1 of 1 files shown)


sarjona (approved):
Hi @timhunt!
Thanks a lot for working on this improvement. I totally agree with all the points you addressed/improved so I'm more than happy to approve it :-)

Besides, it also opens the door to mention here that [Generator tool](https://moodledev.io/general/development/tools/generator) can be used for manual tests, to make easier to setup testing scenario ;-)

PR #1131, opened 2024-09-23
Update 09.md

marcusgreen:
Three of the topics for the presentation


sarjona (approved):
Thanks for updating the agenda, Marcus :-)

PR #1133, opened 2024-09-23
Recording link

wildgirl:


sarjona (approved):
Thanks for updating the meeting recording. However, as you'll see, I've updated it with the specific link for this meeting (for consistency with the rest of the meetings):

meeting_url

PR #1150, opened 2024-10-08
Next release

andrewnicols:
This commit adds the following changes:
- a new version management script to make modifying the versions.json far easier
- a fix for some 1.9 version data
- the next set of release information for each branch

### DIFF
--- a/scripts/version.mjs
+++ b/scripts/version.mjs
@@ -1,0 +1,465 @@
+#!/usr/bin/env node
+/**
+ * Copyright (c) Moodle Pty Ltd.
+ *
+ * Moodle is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Moodle is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* eslint-disable import/no-extraneous-dependencies */
+
+import { program } from 'commander';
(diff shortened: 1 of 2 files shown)


scripts/version.mjs
```diff
@@ -0,0 +1,451 @@
...
+    });
+
+    return major;
+};
+
+const getReleaseDate = async (defaultDate = null) => (
+    await inquirer.prompt({
+        type: 'input',
+        name: 'releaseDate',
+        message: 'What is the expected release date?',
```
junpataleta, line 112: It might be good to specify acceptable date formats.

scripts/version.mjs
```diff
@@ -0,0 +1,451 @@
...
+        let calculatedDate;
+        calculatedDate = new Date(new Date(releaseDate).setMonth(releaseDate.getMonth() + supportLength));
+        calculatedDate = new Date(calculatedDate.setDate(calculatedDate.getDate() - 1));
+        return getFormattedDate(calculatedDate);
+    };
+
+    const { theDate } = await inquirer.prompt({
+        type: 'input',
+        name: 'theDate',
+        message: 'What is the general support end date?',
```
junpataleta, line 213: Should be "security support end date"

junpataleta (changes requested):
Thanks, @andrewnicols! The version script is really handy! 

It just needs some minor changes from my review comments.

I'd also suggest we document this script either as part of
- [Task 9 (release notes task) of the 1 week prior](https://moodledev.io/general/development/process/release#1-week-prior) release tasks
- [Task 4 of the Release day tasks](https://moodledev.io/general/development/process/release#release-day).

What do you think?

andrewnicols:
I've updated with the requested changes and improved the help for majors.

> I'd also suggest we document this script either as part of

I agree

> * [Task 9 (release notes task) of the 1 week prior](https://moodledev.io/general/development/process/release#1-week-prior) release tasks

Usually Mick is the one who does this. @mickhawkins which step do you do this in? We should try and document it.

> * [Task 4 of the Release day tasks](https://moodledev.io/general/development/process/release#release-day).
> 
> What do you think?
That step is already wrong. Again I think it's done by Mick about a day or two before a release when he does the release notes.

junpataleta (approved):
Thanks, Andrew! Looks good!

PR #1159, opened 2024-10-15
Update Inplace editable docs with new events and template

vuvanhieu143:
The current document is outdated since the code have changes.
So it is hard to using the example as references.


andrewnicols (approved):
I've added an additional commit to backport these changes to all supported docs, and make some additional updates and clarifications.

PR #1161, opened 2024-10-18
Add release process task for update moodle-plugin-ci test matrix and templates.

kabalin:
Adding checklist item to remember about bumping moodle-plugin-ci test matrix (as discussed with @junpataleta)


general/development/process/release/index.md
```diff
@@ -168,6 +168,7 @@ Usually on Monday
...
+| 8. | &#10003; | &#10003; | Update test matrix at [moodle-plugin-ci](https://github.com/moodlehq/moodle-plugin-ci/) GHA and Travis tests workflows. For Major release also update `*.dist.yml` templates to reflect current major branch. | Integration Team / moodle-plugin-ci maintainer |
```
junpataleta, line 171: Hi, @kabalin. Do we need to perform this task during minor releases, too?

junpataleta (commented):
I'm not sure if we need to perform this task for the minor release cycle as well, so I'd like to ask for some clarification and pros/cons of doing it during minor releases as well instead of only major releases.

general/development/process/release/index.md
kabalin: Hi Jun, Eloy bumped test matrix in the past after minor releases, e.g. https://github.com/moodlehq/moodle-plugin-ci/commit/bce47eda9917bd5557686518a4d37224399ccfaf, but at the same time it is not critical as this is just a test matrix update, I would classify it as post-release housekeeping task (it will likely pass as main is always tested and any problem would be spotted earlier). On the other hand, I don't know if this type of change fits release checklist, thus the proposal to include it, so leaving to you to accept only major change part of it or both :)

junpataleta: Thanks, @kabalin. The addition of the 404 branch to the test matrix seems to have been done by Eloy when 4.4 was released, so it coincides with the major release cycle. Thus, perhaps adding this task only for major releases is sufficient. Thanks!

junpataleta (approved):
Thanks! Looks good to go now!

PR #1162, opened 2024-10-21
Fork documentation for Moodle 4.5

junpataleta:
Part of the [2 weeks after the release process ](https://moodledev.io/general/development/process/release#2-weeks-after)


andrewnicols:
Note: I haven’t done this because I wanted to check if we should remove the 4.2 docs first to avoid excessive documentation bloat

junpataleta:
Hi, @andrewnicols. Can't we deal with that on a separate issue instead?

andrewnicols:
We can but then we are building more docs for longer which makes the build
bigger and slower. That’s the only reason I was holding off.

It it’s building fine then we an land this now :)

Andrew Lyons
Principal Architect

Moodle HQ - moodle.com [image: Moodle - Empowering educators to improve our
world] <https://moodlemoot.org>
Join the Moodle Community at #MootGlobal24. Mérida, México, 22-24 October
2024

HuongNV13 (approved):

HuongNV13:
Thanks, Jun.
It looks good to me.
Since this PR was built successfully, I guess we can land this and create another PR to remove the 4.2's devdocs

PR #1168, opened 2024-10-25
Fix the intro page for the 5.0 dev docs

junpataleta:
I thought I had already sorted out this in #1162, but apparently, I missed it. 

I also added another commit that fixes the wording for future non-LTS versions in the release state summary.

### DIFF
--- a/docs/intro.md
+++ b/docs/intro.md
@@ -1,13 +1,13 @@
 ---
 id: introduction
 title: Introduction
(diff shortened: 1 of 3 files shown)


junpataleta:
I added another commit that updates the `general/documentation/forking-versions.md` and ensure that updating `docs/intro.md` does not get missed.

andrewnicols (approved):

PR #1174, opened 2024-11-04
Removal of docs for Moodle 4.2

andrewnicols:
Docs have been preserved by triggering a build immediately before this change and using its permalink.


andrewnicols:
Self merging after a week.

PR #1181, opened 2024-11-11
fix #1180 Update index.md $plugin->incompatible

### LINKED ISSUE #1180: [docs] Error in version.php

https://moodledev.io/docs/5.0/apis/commonfiles/version.php

The example for $plugin->incompatible is using an array, which is wrong, and should be an integer

Yes - I have one prepared

gbarat87:
$plugin->incompatible = [401];
Wrongly defined in the example
It require an Integer and not an array


andrewnicols (approved):
Thanks for fixing that issue. I have taken the liberty of porting this to other versions of this file.

PR #1188, opened 2024-11-12
Add meeting recording link

wildgirl:


sarjona (approved):
Thanks a lot Helen! As discussed, I also added the link to the presentation :-)

PR #1190, opened 2024-11-13
A few dates are now in the past (April and October 2024)

stronk7:
Not much to say, just those dates have already passed.


sarjona (approved):
Well spotted, Eloy! :-)

Issue #1192, opened 2024-11-14
[docs] Create documentation for implementing report builder support in plugins [Feature]


NeillM:
### What is the Moodle feature that needs documenting?

The [report builder API documentation](https://moodledev.io/docs/5.0/apis/core/reportbuilder) seems to talk about how it is implemented and some of the things that can be done with it, especially in regards to implementing system reports with it.

It is however entirely unclear to me from the documentation what a plugin would need to do to add support, for example what must be added to a plugin and where it must be placed in a minimum implementation.

As an example the [attendance plugin extends a core_reportbuilder\datasource](https://github.com/danmarsden/moodle-mod_attendance/blob/MOODLE_404_STABLE/classes/reportbuilder/datasource/attendance.php#L34) which is not mentioned in the documentation at all

### Is this documentation specific to a Moodle version?

None

### Are you able to provide a patch for this?

None

NeillM:
I'm currently working on improving them in: https://github.com/NeillM/devdocs/tree/reportbuilder

So far I have

- found that the 4.4 versions of the docs seem to not have been updated when some parts of the API were changed
- I have added a section about how to create a data source for custom reports
- Added some additional details about some of the things I have found out about them while implementing custom reports for one of our plugins

andrewnicols:
Pinging @paulholden FYI

andrewnicols:
Thanks for working on this @NeillM :)

NeillM:
Since I'm having my first foray into using the report builder API I might as well.

When you get your head around what you need to do it is a very nice system.

NeillM:
I have also now added some docs to explain how to:

- Add unit tests for `datasources`
- Details about the things you need to do to define conditions for each filter type (so that others do not have to trawl through the code)

PR #1196, opened 2024-12-02
Update report builder API docs

NeillM:
I have:

- Added documentation for creating data sources for custom reports
- Setting filter conditions in code
- Added some additional tips for creating entities
- Pointed code examples at the branch of Moodle the docs are for

I have attempted to ensure that the docs include API changes that have happened since Moodle 4.3

This change should hopefully implement most of the improvements I think would be useful in Issue #1192

### DIFF
--- a/docs/apis/core/reportbuilder/filtering.md
+++ b/docs/apis/core/reportbuilder/filtering.md
@@ -1,0 +1,329 @@
+---
+title: Setting default conditions
+tags:
+  - Report builder
+  - reports
+---
+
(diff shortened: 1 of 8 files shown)


NeillM:
I think that the issue is that on Windows if you run `yarn lint-staged` on something like `./filter.md` it will be converted to a path like `\docs\apis\core\reportbuilder\filtering.md`, when in fact it should be `/docs/apis/core/reportbuilder/filtering.md`

paulholden:
Thanks @NeillM for documenting this - I'll have a look through these

paulholden (approved):

NeillM:
Thanks Paul

Issue #1197, opened 2024-12-02
[repo] Running yarn lint-staged on Windows with changes containing links causes pull request build failures


NeillM:
### What is the issue that you're experiencing?

On Windows if you run `yarn lint-staged` on a commit that contains a link like `[my link](./filter.md)` it will be converted to a path like  `[my link](\docs\apis\core\reportbuilder\filtering.md)`

I'm guessing it should probably be just left alone, since changing it to `[my link](/docs/apis/core/reportbuilder/filtering.md)` causes it's own issues.

### Are you able to provide a patch for this?

Yes, in pull request #1198

NeillM:
It looks as though it is `markdownlint-cli2 --fix` that is doing this.

NeillM:
It looks as though we are using the [node.js path library](https://nodejs.org/api/path.html) in a way that makes the links work dependant on the OS, we can force it to always use more Linux like links by changing from just `path` to `path.posix`

PR #1198, opened 2024-12-03
Fix link checker on Windows

NeillM:
Before this change the way links formed differed based on operating system.

This meant that on Windows the fix-renamed links checker would both change links that it should not, and leave them with Windows style slashes.

This is a fix for Issue #1197


andrewnicols (approved):
Thanks for this. Looks good to me.

PR #1199, opened 2024-12-03
Add question bank filter API guide

marxjohnson:
This functionality was added in MDL-72321. MDL-78220 will add several new filters, so this guide may expand as the children of that tracker are developed.

### DIFF
--- a/docs/apis/plugintypes/qbank/filters.md
+++ b/docs/apis/plugintypes/qbank/filters.md
@@ -9,20 +9,17 @@ documentationDraft: true
 ---
 
 <Since
-version="4.3"
-issueNumber="MDL-72321"
+    version="4.3"
+    issueNumber="MDL-72321"
 />
 
-Question bank plugins allow you define additional filters. These can be used when viewing the question bank, and are included
-in the URL so that a filtered view of the question bank can be shared. They are also used when defining the criteria for adding
-random questions to a quiz.
+Question bank plugins allow you define additional filters. These can be used when viewing the question bank, and are included in the URL so that a filtered view of the question bank can be shared. They are also used when defining the criteria for adding random questions to a quiz.
 
 ## Creating a new filter condition
 
 A filter condition consists of two parts - the backend "condition" PHP class, and the frontend "filter" JavaScript class.
 
-The "condition" class defines the general properties of the filter - its name, various options, and how it is applied to the
-question bank query.
+The "condition" class defines the general properties of the filter - its name, various options, and how it is applied to the question bank query.
 The "filter" class defines how the filter is displayed in the UI, and how values selected in the UI are passed back to the condition.
 
 Each new filter condition must define a new "condition" class in the qbank plugin based on `core_question\local\bank\condition`.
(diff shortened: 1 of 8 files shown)


timhunt (approved):
Thanks Mark. This looks like a good start to documenting this.

timhunt:
Ah! but I see taht the automated checks need to be appeased.

docs/apis/plugintypes/qbank/filters.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Basic example
+
+This outlines the bare minimum required to implement a new filter condition. This will allow you to filter based on a pre-defined
+list of values, selected from an autocomplete field. This assumes that you already have the basic framework of a qbank
+plugin in place. For real-world examples, look for classes that extend `core_question\local\bank\condition`.
+
+Create a `condition` class within your plugin's namespace. For a plugin called `qbank_myplugin` this would look something like:
+
+```php
```
andrewnicols, line 39: I find it's really handy to specify the location of the class in the title, for example:
```suggestion
```php title="question/bank/myplugin/classes/myfilter_condition.php"
```

docs/apis/plugintypes/qbank/filters.md
```diff
@@ -0,0 +1,215 @@
...
+
+}
+```
+
+Define the `get_name` method, which returns the label displayed in the filter UI.
+
+```php
+    public function get_name(): string {
+        return get_string('myfilter_name', 'myplugin');
+    }
```
andrewnicols, line 54: Since this example is in isolation, you could either:
* add the class wrapper to make it clearer; or
* remove the indent to make it look cleaner in the built code.

```suggestion
```php
class myfilter_condition extends condition {
    public function get_name(): string {
        return get_string('myfilter_name', 'myplugin');
    }
}
```


Or

```suggestion
```php
public function get_name(): string {
    return get_string('myfilter_name', 'myplugin');
}
```

andrewnicols: You may also want to use the title attribute here, but that's up to you. If you do I wouldn't set it to the filename, but a short sumarry of what you're doing, for example:

```
title="Define the filter name"

File: docs/apis/plugintypes/qbank/filters.md [hunk 3]
```diff
@@ -0,0 +1,215 @@
...
+namespace qbank_myplugin;
+
+use core_question\local\bank\condition;
+
+class myfilter_condition extends condition {
+
+}
+```
+
+Define the `get_name` method, which returns the label displayed in the filter UI.
```
andrewnicols, line 49: You may also want to consider adding parens to method names:
```suggestion
Define the `get_name()` method, which returns the label displayed in the filter UI.
```

Your call but if you do please do so for all of them.

docs/apis/plugintypes/qbank/filters.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Filter options
+
+If your condition supports additional options as to how the selected values are applied to the query, such as whether child
+categories are included when parent categories are selected, you can define "Filter options".
+
+In your condition class, define `get_filteroptions()` which returns an object containing the current filter options. You will
+probably want to add some code to the constructor to read in the current filter options, and some code the `build_query_from_filter`
+to use the option. See
+[`qbank_managecategories\category_condition`](https://github.com/moodle/moodle/blob/main/question/bank/managecategories/classes/category_condition.php#L331)
```
andrewnicols, line 200: Consider making this link point to a specific commit hash, or to just the file as a whole?

andrewnicols (changes requested):
These docs look generally great. I've left a few comments -- see what you think.

Since these are docs for a 4.3 feature we should backport to 4.3 and 4.4 using the `versioned_docs` folders. If you need help with this I can do this on the PR -- just let me know.

Thanks

docs/apis/plugintypes/qbank/filters.md
marxjohnson: Done

marxjohnson:
@andrewnicols Thanks for your comments, I have made those changes and re-worked the ordering of things a bit to make it clearer which bits are compulsory, and which are more advanced. I've included those changes as fixups so I can squash them if you're happy.

I've also added an extra commit to cover the validation support I've added in MDL-83862.

marxjohnson:
@andrewnicols Could you do that `versioned_docs` thing you mentioned? I've not done that before I'd be grateful of the help!

andrewnicols (approved):

marxjohnson:
Now this has been merged, should it be showing up on the live site or is there there a "release" that we need to wait for? I don't see it here: https://moodledev.io/docs/5.0/apis/plugintypes/qbank

PR #1200, opened 2024-12-03
Change discovery_agent interface name

volodymyrdovhan:
Interface '\core\hook\hook_discovery_agent' does not exist in the \core\hook, it should be '\core\hook\discovery_agent'.

image


andrewnicols (approved):
Good catch on this one.

PR #1206, opened 2024-12-11
Add get_fieldset()

leonstr:
Presumably this function is missing and wasn't deliberately omitted?

Can this change be cherry-picked to the 4.4 and 4.5 docs or should these be changed in separate requests?


andrewnicols (approved):
Thanks for adding this.

Please note that this is not meant to be an exhaustive list of methods. Please refer to the phpdoc for this.

PR #1213, opened 2024-12-19
Re-order the peer review checklist into a more logical order

timhunt:
At least, this seems more logical to me:

* Move the checkes that are more overall to the end.
* Group all the UI, and DB/performance together.

Also, add mention of developer_notes label which I always forget.


timhunt:
If this is accespted, then we shold updated the Peer review canned response in the tracker to match.

sarjona (approved):
Hi @timhunt!
Thanks for sharing this proposal. I agree the order makes more sense and it's also great you improved docs. Well done! <3

@wildgirl could you help us upgrading the canned response in the Tracker with the new order? :-)

PR #1217, opened 2024-12-20
Display banner for versioned docs

sarjona:
William Entriken requested this at https://tracker.moodle.org/browse/MDLSITE-7820. To implement it, we're taking advantage of the banner setting: https://docusaurus.io/docs/versioning#configuring-versioning-behavior.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -28,7 +28,7 @@ dotEnvConfig();
 
 const versionConfig = Object.fromEntries(Versions.map((version) => [version, {
     label: version,
-    banner: 'none',
+    banner: 'unmaintained',
 }]));
 versionConfig.current = {
     label: `main (${nextVersion.nextVersion})`,


andrewnicols:
Can you give an example of where you want to use this?

we already have some custom version-specific banners

andrewnicols:
I’d prefer to do this in a different test. Check out how it looks sit. This patch on 4.1 for example

sarjona:
Hi @andrewnicols!
I'm not sure if I understand your question. The idea of this patch is to implement what William Entriken requested at https://tracker.moodle.org/browse/MDLSITE-7820: 

`Add a warning that you are reading version 4.4 of the docs but a newer version is available. https://moodledev.io/docs/4.4/apis/commonfiles/version.php`

Docusaurus implements this with the banner setting, so the patch is enabling this, to display a banner at the top of all the versioned doc pages:
MDLSITE-7820_banner

andrewnicols:
Compare against the message here: https://moodledev.io/general/releases/4.1

this is what I want to show as it has version-specific information and is factually incorrect.

it is not correct to say that 4.5 is “no longer actively maintained”.

If I recall the existing banner shown here is meant to be shown on all pages of versioned docs and something has broken that. The Docusaurus banner is too limiting.

sarjona:
I'm closing this because a different approach has been prepared in https://github.com/moodle/devdocs/pull/1219 (although it's unclear if it will be integrated).

PR #1219, opened 2024-12-22
[repo] Show a warning for stable versions that newer docs are available.

andrewnicols:
[~sarjona]: Here's the alternative for #1217.

I 100% do not agree with doing this and I have no intention of landing this. It's a blight on the documentation and I feel that it serves no useful purpose but to add more junk on the top of the page and to push the _useful_ part of the docs out of the immediate code window.

The pages you want to add this to are for fully supported versions of the codebase. I also do not agree that we should point people to docs for _unstable_ versions of the codebase.

I won't land this.

### DIFF
--- a/nextVersion.js
+++ b/nextVersion.js
@@ -15,11 +15,13 @@
  * along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+const latestVersion = '4.5';
 const nextVersion = '5.0';
 const nextLTSVersion = '5.3';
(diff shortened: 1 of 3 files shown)


sarjona:
Hi @andrewnicols!

Thank you for working on this PoC to implement William Entriken's request at https://tracker.moodle.org/browse/MDLSITE-7820 :-) 

I understand your point, but I also see the benefits of displaying this banner for older versions. Perhaps we could consult the UX team for a less intrusive way to display it... Thoughts?

PR #1229, opened 2025-01-10
Fix typo

micaherne:
There is a typo in the example code for resetting the container - a colon at the end instead of a semi-colon.


sarjona (approved):
Hi @micaherne!
Thanks for reporting and fixing this. Well spotted! As you'll see, I merged this with https://github.com/moodle/devdocs/pull/1230 and also added docs/apis/core/di/index.md to guarantee it's fixed everywhere.

PR #1234, opened 2025-01-16
[docs] Add PHPUnit 11 upgrade guidance

andrewnicols:
Note: Whilst this PR is associated with MDL-83468 I do not think we should block it until that issue lands.

This guidanc applies to the current version of PHPUnit too and will help people to upgrade before we make the change.


general/development/tools/phpunit/upgrading-11.md
```diff
@@ -0,0 +1,198 @@
...
+
+You may need to wrap your asserted value in `array_values()` if the keys are not important.
+
+### The `setMethods` method on the `MockBuilder` is deprecated and removed
+
+#### Problem
+
+The `setMethods` method was deprecated in PHPUnit 8.3.0 and removed in PHPUnit 10.
+
+See https://github.com/sebastianbergmann/phpunit/issues/3687 for further information on the rationale.
```
junpataleta, line 101: Would be good to make a readable link for accessibility. For example,

```
See [PHPUnit Issue #3687](https://github.com/sebastianbergmann/phpunit/issues/3687) for further information on the rationale.
```

general/development/tools/phpunit/upgrading-11.md
```diff
@@ -0,0 +1,198 @@
...
+
+Many of the features removed in PHPUnit 10 were then partially added back in PHPUnit 11.
+
+To give a specific example, PHPUnit 9.6 has a `--debug` flag. This was removed in 10.0, and added back in 11.0 as an alias for a combination of other flags.
+
+There are lots of other examples of this kind of thing.
+
+## Can I make my tests work in both 9.6 and 11.4?
+
+Yes, _usually_. But it may require some changes to your tests and/or code. For the most part any such changes are a good thing and in all cases I've encountered they've been because the tests were wrong in the first place anyway.
```
junpataleta, line 35: Would the "I" in `I've encountered` sound confusing to the reader, given that authors are not indicated in the dev docs pages (only the person who last updated the page)? Would something like `we've encountered` be more appropriate?

general/development/tools/phpunit/upgrading-11.md
```diff
@@ -0,0 +1,198 @@
...
+
+Data providers _must_ now provide data. You cannot have data providers which return an empty Iterator/Array.
+
+```php
+/**
+ * @requires extension Redis
+ */
+```
+
+#### Solution - use the `@requires` annotation
```
junpataleta, line 63: It's not quite clear to me how using `@requires` works or how to use it to fix the empty data provider issue. Is it being added to PHPDoc block of the test function that requires the data provider?

Maybe adding an example of an empty data provider with the corresponding `@requires` would be good?

general/development/tools/phpunit/upgrading-11.md
```diff
@@ -0,0 +1,198 @@
...
+
+You may be able to use the [`@requires`](https://docs.phpunit.de/en/9.6/annotations.html#requires) annotation to skip your test based on some configuration, for example whether a specific PHP Extension is installed, for example:
```
junpataleta, line 65: The PHPUnit doc link for the `@requires` annotation points to the 9.6 doc. Should this be set to 11.4 instead?

general/development/tools/phpunit/upgrading-11.md
```diff
@@ -0,0 +1,198 @@
...
+            'baz' => 'qux',
+        ],
+    ]''
+}
+```
+
+Then the test that uses this data provider must have the same parameter names:
+
+```php
+/**
```
junpataleta, line 181: Super minor PHPDoc block edit suggestion:

```php
/**
 *
 * @dataProvider my_provider
 */
```

junpataleta (changes requested):
Thanks, @andrewnicols! This looks almost ready! I just have some minor edit suggestions for your consideration.

Cheers!

general/development/tools/phpunit/upgrading-11.md
andrewnicols: There's literally an example just above. I'll move it to the solution. We can't do _everything_. People need to take some responsibility and accountability for themselves.

general/development/tools/phpunit/upgrading-11.md
andrewnicols: Fixed.

junpataleta (approved):
Thanks, Andrew! Looking good!

PR #1245, opened 2025-01-29
[docs] Add basic rules for routes

andrewnicols:
This is an initial set of rules for routes.

I've tried to keep things relatively simple but I may have dumbed it down too much.


kabalin:
Rendered page: https://deploy-preview-1245--moodledevdocs.netlify.app/general/development/policies/routing

general/development/policies/routing/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+:::danger Default and non-default entities
+
+The use of both a _default_ entity _and_ a non-default entity is not officially supported.
+
+:::
+
+:::tip Singular or Plural
+
+Where present, entities should always be in the plural form.
```
kabalin, line 74: I agree about [plural form](https://tracker.moodle.org/browse/MDL-76834?focusedId=1122537&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1122537)! Should examples for course entity be changed accordingly?

andrewnicols:
The issue with the course entity is that it is the name of the subsystem
itself. I’m not sure how we would pluralise a subsystem name

general/development/policies/routing/index.md
mattporritt: We don't need to strictly follow the subsystems names here. 
However, in the example of course, `/courses/<courseid>/view` sounds not correct, as you are only going to ever look at a singular course. So I think in this case singular is the way to go. 
Not sure how we make the docs clearer though.

andrewnicols: At the moment the component resolution is automatic based on the location of the class and we automatically normalise it.

If we want to do something different then we'll have to work out the best way of doing so and in which situations.

kabalin: > However, in the example of course, `/courses/<courseid>/view` sounds not correct, as you are only going to ever look at a singular course. So I think in this case singular is the way to go.

Yeah, it is sounds not correct in individual scenario, unless one see it as query "from courses select <courseid>". Thinking in broader context, what will be the endpoint for querying list of courses? Different `/courses` path or `/course/list` action? Something better to consider while we are at early stage I guess :)

With regard to this "tip", to avoid contradition for now we might say "Where possible, entities should always be in the plural form."

PS In Martin Fowler [REST maturity model](https://martinfowler.com/articles/richardsonMaturityModel.html) article all entities are plural.

mattporritt: Andrew's point is a good one that resolution is based on automatic resolution. Messing around with that feels like it may add more complexity than it's worth.
Also, thanks for the extract context Ruslan

PR #1248, opened 2025-01-31
[repo] Update Husky installation for Yarn2

andrewnicols:
Yarn 2 does not run the `prepare` script. We need to switch to `postinstall` or `install`. Husky recommends using the postinstall.


snake (approved):
This change looks good to me. I've run the usual setup (nvm install, npm install yarn, yarn) and can confirm that the .git/config has the following line added (which was missing before):
hooksPath = .husky

Happy to move this one forward.

PR #1251, opened 2025-02-03
[docs] Add notes about plugin removal from core

andrewnicols:
This PR starts to formulate a process for removing plugins from core.

### DIFF
--- a/general/development/policies/deprecation/plugin-removal.md
+++ b/general/development/policies/deprecation/plugin-removal.md
@@ -1,0 +1,97 @@
+---
+title: Plugin Removal
+tags:
+  - Processes
+  - Core development
+  - Deprecation
+  - Plugins
+  - Removal
+---
+
+We regularly assess the standard plugins that are included in Moodle and will, from time-to-time, remove some of these.
+
+When this happens we will announce to the community that this will be happening.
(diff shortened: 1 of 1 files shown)


andrewnicols:
Pinging @danmarsden

mattporritt:
@andrewnicols 
I think we need to be a bit clearer in the annoucment about the plugin being remove. We may want to say something that we will let the community know via the moodle.org forums and the tracker before a plugin is removed from core. And will endeavour where POSSIBLE to allow sufficent time for those using the plugin to understand the change and to update any workflows or dependencies.

danmarsden:
Thanks @andrewnicols - this looks much more sensible to me than the previous approach - I really like the statement "Plugins without a maintainer will not be accepted into the plugins database." and I suspect we should put this into the plugins db checklist as well (pinging @mudrd8mz to hear your thoughts too!)

andrewnicols:
Thanks both,

@mattporritt I've added to the docs to list of the specific places we aim to announce, and to put a time disclaimer in.

There are some cases where we won't announce but it probably isn't worth noting them. One such example is h5plib.

mattporritt (approved):

PR #1246, opened 2025-01-30
[docs] replace old docs links with their new counterparts

ziegenberg:
The reply templates for pasting into the tracker still had the old links to docs.moodle.org/dev.

PR #1263, opened 2025-02-16
[docs] fix links to devdocs

ziegenberg:
In #1246 I was made aware by @andrewnicols that I placed the wrong links. Let's fix this.


sarjona:
Hi @ziegenberg!
Thanks for taking care of this!  Just a heads-up: the only spot that needs the change is ../../tools/gha.md. The other files are canning templates, so it's best to keep the full URL there for anyone copying them.
I've updated the commit, so I'm good to merge it :-)

sarjona (approved):

PR #1268, opened 2025-02-24
[docs] Add information about new human date renderers

sarjona:
As part of MDL-83873 (https://tracker.moodle.org/browse/MDL-83873), new human date renderers were added. This commit provides the necessary documentation for these renderers.


ferranrecio:
I believe that mentioning the library solely in the upgrade notes is insufficient. Over time, this documentation may become overlooked; therefore, the component deserves its own dedicated page. I've included a fixup commit that you can use for reference (first time I use gh command, hope it goes as expected).

PR #1274, opened 2025-02-26
MDL-83888 new overview redirection and renderer helper

ferranrecio:
This pull request is related to the patch on [MDL-83888](https://tracker.moodle.org/browse/MDL-83888).

### DIFF
--- a/docs/apis/plugintypes/mod/courseoverview.md
+++ b/docs/apis/plugintypes/mod/courseoverview.md
@@ -202,6 +202,38 @@ class overview extends activityoverviewbase {
 }
 ```
 
+### Custom grade overview items
+
+If the activity has a grade item, the course overview will display the grade of the activity to the student. If your plugin has one single grade item, you don't need to do anything.
+
+For activities with more than one grade item (for example, mod_workshop), no grade will be shown by default because the system doesn't know the meaning of each grade item. For those cases, the plugin must implement `get_grade_item_names` to provide the generic name for each grade item. The overview report will show only the grade items with a generic name.
+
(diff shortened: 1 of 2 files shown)


docs/apis/plugintypes/mod/courseoverview.md
```diff
@@ -202,6 +202,38 @@ class overview extends activityoverviewbase {
...
+        // Names must be indexed by the grade item id.
+        $names[$item->id] = get_string($stridentifier, 'mod_YOURPLUGIN');
+    }
+    return $names;
+}
+```
+
+:::note
+
+It is not recommended to override the `get_grades_overviews` method. The method is used to provide the grade information to the course overview, but it is not intended to be used to provide extra information. If you don't want to show the grade information in your plugin, you can override the `get_grade_item_names` method and return an empty array.
```
sarjona, line 233: This is not consistent with the PHPDoc displayed for the `get_grades_overviews()` method:

> Plugins can override this method to provide their own grede overview items if necessary.

sarjona (changes requested):
Hi @ferranrecio!
Thanks for working on the devdocs for https://tracker.moodle.org/browse/MDL-83888 :-)

They look good but, as you'll see, there are some inconsistencies between them and the code that should be reviewed...

docs/apis/plugintypes/mod/courseoverview.md
ferranrecio: You are right. I edited the patch PHPDoc to make ti consistent with the documentation. Nice catch.

PR #1280, opened 2025-03-06
MDL-83541 Add qtype restore code documentation

marxjohnson:
This describes the required updates to qtype plugins following to implementation of MDL-83541 to avoid duplication of questions during restores.


timhunt:
Thanks Mark. That looks like a really good new docs page.

The only bit I don't really like is the use of $name for the plugin name. Is this a convention used elsewhere in the docs? I managed to avoid it comletely on https://moodledev.io/docs/5.0/apis/plugintypes/qtype/newquestiondefaults. On https://moodledev.io/docs/5.0/apis/plugintypes/qbank/filters we use qbank_myplugin. Up to you to make a final call.

andrewnicols:
I'd also second not making this a var. Another option would be to use a placeholder:

```md
In your  `restore_qtype_<name>_plugin`  class, add an override of the `remove_excluded_question_data()` method, which removes this
```

timhunt:
I like Andrew's suggestion. Feel free to updated [qbank/filters](https://moodledev.io/docs/5.0/apis/plugintypes/qbank/filters) while you are at it Mark (he suggested cheekily :-))

HuongNV13 (approved):

HuongNV13:
Thanks, Mark
It looks good to me

HuongNV13:
I have rebased the branch, and the failures are gone!

PR #1282, opened 2025-03-10
Adding Marie to agenda

wildgirl:


sarjona (approved):
Thanks Helen!
I replaced directional ` with ' in the agenda and also fixed a couple of unrelated issues to make LINT job happy :-)

PR #1286, opened 2025-03-11
Add missing Atto plugintype page for 4.1

sarjona:
It's causing some failures when running (because this page was not backported when it was created)


sarjona:
I'm self-merging this because this failure is causing devdocs are not upgraded with the latest patches merged :-)

lameze (approved):
Looks good Sara

PR #1288, opened 2025-03-11
Fix broken formatting

leonstr:


andrewnicols (approved):

andrewnicols:
Thanks Leon,

I've applied the same patch to all other versions of the docs.

PR #1293, opened 2025-03-13
Declare PHP 8.4 supported

andrewnicols:
- Do it in the PHP policy page: By MDL-80117, since 5.0
- Update 5.0 release notes (not needed in this case, it's already showing the correct PHP 8.4 information).


junpataleta (approved):
Yay!

PR #1295, opened 2025-03-13
[Docs] Adding new Writing for Accessibility page

juliavs85:
Adding a new page to the Product Writing Guidelines.


timhunt:
Thanks for doing this. Looks good. A few suggestions:

* "Learn more about all our plans" in a 'good' example. I feel like this would be an even better example if you removed the word "all".
* The "HTML page titles" contracticts how Moodle currently works and good practice. "Manage activities" is not a good page title. (Suppose you are trying to compare things between, say Moodle academy and Moodle.org. It should be something like "Manage activities | Course: Set up your Moodle Development Environment | Moodle Academy" - which is pretty much what Moodle currently does. Oh! your later example in this section say that. I think it is bad to start with some misleading 'Good' examples.

timhunt:
Those chagnes look good to me. THanks @juliavs85.

timhunt (approved):
This looks good to me. Am I OK to add an approving review?

PR #1302, opened 2025-03-20
Final deprecation policy update to keep method signature

abgreeve:
The documentation for deprecation needed clarifying for how to handle the final deprecation process.
As Moodle code is more OOP we can't follow the existing final deprecation steps.

### DIFF
--- a/general/development/policies/deprecation/index.md
+++ b/general/development/policies/deprecation/index.md
@@ -63,9 +63,11 @@ debugging('foobar() is deprecated. Please use foobar::blah() instead.', DEBUG_DE
 <TabItem value="core_deprecation" label="Using the \core\deprecation API from Moodle 4.4">
 
 ```php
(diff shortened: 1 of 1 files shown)


general/development/policies/deprecation/index.md
```diff
@@ -135,16 +137,18 @@ throw new coding_exception(
...
+    return [];
 }
 ```
 
 </TabItem>
 
 </Tabs>
 
-- All function parameters should be removed.
+- When deprecating methods in a class, retain the existing parameters and return type.
```
andrewnicols, line 150: Honestly I’d just simplify it further. Make it the same for class methods, and functions. There is no benefit in changing the signature so just leave the whole thing

abgreeve: Okay, will change it over.

andrewnicols (approved):

PR #1306, opened 2025-03-24
Clarify precautions around hooks which dispatch on install/upgrade

cameron1729:


docs/apis/core/hooks/index.md
```diff
@@ -296,9 +296,10 @@ Any guidelines for callback priority should be described in hook descriptions if
 
 :::caution
 
-Callbacks _are executed during system installation and all upgrades_, the callback
-methods must verify the plugin is in correct state. Often the easiest way is to
-use function during_initial_install() or version string from the plugin configuration.
+Unlike lib.php based plugin callbacks, certain hooks _are dispatched during system installation and upgrade_ (for example `before_http_headers`). Callback
+methods for such hooks must take extra care to ensure the plugin is properly initialised, as (among other things) on install _the site's database is not initialised_.
+The function during_initial_install() or version string from the plugin configuration may be used to conditionally make DB queries or use API functions in such cases.
+Failure to do so may render the web install/upgrade page unusable.
```
andrewnicols, line 302: No need to linebreak - it makes diffs shit and has no impact.

Can you please put the methods, filepaths, etc. in backticks while you're at it.

I'd also suggest you don't need to say that "certain hooks" - all hooks may be. There is no definitive list. It just depends whether the hook is triggered during the process.

Perhaps something like:

Hooks may be dispatched at any time, including during _installation and upgrade_, therefore the callback methods for hooks which may be triggered during installation and upgrade must take care to ensure that the plugin is properly initialised and that the database is available if database calls are made.

The `during_initial_install()` function can be used to check whether the Moodle site is installed, and fetching the plugin's version using `get_config('your_pluginname', 'version')` are two ways that this can be achieved.  Failure to do this may render the web installation and upgrade system unusable.

Please note that the legacy component callback system where callbacks were created as global functions in `lib.php` could _not_ be called during installation or upgrade.

cameron1729:
Thanks @andrewnicols - I've updated the wording, removed line breaks, and added backticks.

andrewnicols (approved):

PR #1307, opened 2025-03-25
[docs] Link to March meeting

andrewnicols:


sarjona (approved):
Hi @andrewnicols!
I think this is usually done after the meeting but I agree it will probably help people to find it (so I'm happy to approve and merge it).

PR #1308, opened 2025-03-25
recording link, removing check this time in your location link

wildgirl:


sarjona (approved):
Thanks Helen! I removed one trailing space after UTC which was causing a lint failure :-)

PR #1312, opened 2025-03-31
[docs] Further clarify final deprecation procedure

andrewnicols:


paulholden (changes requested):
I think this is unnecessary in strictness regarding functions - should we remove that reference and just state "class methods"?

We also appear to be missing any reasoning, perhaps we should mention this specifically refers to class methods for which there is a reasonable expectation they may have been overridden in child classes

paulholden (approved):

PR #1313, opened 2025-04-01
[docs] Clarify meaning of public for deprecation policies

andrewnicols:


paulholden (approved):
Thanks Andrew, good clarification!

PR #1326, opened 2025-04-12
Fix typo in code example

hribeiruporto:
parameters list with an dot instead of comma


sarjona (approved):
Hi @hribeiruporto!
Well spotted! You're right! :-)
The patch looks great, and I also backported it to the other versions (not just 4.5 but 4.1, 4.3, 4.4 and main). Good job!!

PR #1336, opened 2025-04-25
Changing the Upgrading link so it goes to the 4.5 docs

wildgirl:


andrewnicols (approved):
Thanks for that - I'll add it to my fork notes for 500.

andrewnicols (approved):

PR #1345, opened 2025-05-05
5.x Release dates

andrewnicols:
This changset also adds the ability to hide dates so that we can have them in the calendar but not visible just yet.

### DIFF
--- a/data/versions.json
+++ b/data/versions.json
@@ -2,6 +2,40 @@
     "$schema": "../static/schema/versions.json",
     "versions": [
         {
+            "name": "5.3",
+            "releaseDate": "5 October 2026",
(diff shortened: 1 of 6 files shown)


abgreeve:
Hi Andrew, 
All of the dates for the releases look accurate. Updating the security date for 5.1 should probably have a wider conversation and separate issue. 

I haven't done a pull request for docs in ages and this says there is a conflict with versions.json. Can you let me know if / when this is okay to merge?

abgreeve (approved):
Comment already made.

andrewnicols:
Should be good to merge. No conflicts (andi t was able to self-rebase). You shoudl be able to just select "Merge when ready"

PR #1348, opened 2025-05-07
Update 4.5.md bumping MySQL version

leonstr:
It looks like Moodle won't install with MySQL 8.0.0 because `role` was a reserved word, and is used by `mdl_course_completion_criteria.role`.

This restriction was removed in MySQL 8.0.1.

See https://dev.mysql.com/doc/mysqld-version-reference/en/keywords-8-0.html#keywords-8-0-detailed-R and https://moodle.org/mod/forum/discuss.php?d=467863 for more details.

### DIFF
--- a/general/releases/4.5.md
+++ b/general/releases/4.5.md
@@ -31,7 +31,7 @@ Moodle supports the following database servers. Again, version numbers are just 
 | Database | Minimum version | Recommended |
 | --- | --- | --- |
 | [PostgreSQL](http://www.postgresql.org/) | 13 (increased since Moodle 4.1) | Latest |
-| [MySQL](http://www.mysql.com/) | 8.0 (increased since Moodle 4.1) | Latest |
+| [MySQL](http://www.mysql.com/) | 8.0.1 (increased since Moodle 4.1) | Latest |
 | [MariaDB](https://mariadb.org/) | 10.6.7 (increased since Moodle 4.1) | Latest |
 | [Microsoft SQL Server](http://www.microsoft.com/en-us/server-cloud/products/sql-server/) | 2017 | Latest |
 | [Oracle Database](http://www.oracle.com/us/products/database/overview/index.html) | 19c (increased since Moodle 4.3) | Latest |


andrewnicols:
Hi @leonstr ,

Thanks for raising this. This documentation matches the information held within Moodle in the `admin/environment.xml` file. If this is a necessary change to the requirements then it _must_ be in Moodle too.

It is worth noting that MySQL 8.0.1 was released in 2016 and is not supported. I would not expect anyone to be trying to install Moodle with 8.0.0 anyway.

leonstr:
@andrewnicols _… and is not supported. I would not expect anyone to be trying to install Moodle with 8.0.0_ – The documentation and `environment.xml` indicate that 8.0.0 is supported by Moodle and, as the forum link shows, assuming no one will be using this isn't valid.

I've opened [MDL-85485 Require MySQL 8.0.1 in environment.xml for Moodle 4.5](https://tracker.moodle.org/browse/MDL-85485) for consideration.

leonstr:
_I would not expect anyone to be trying to install Moodle with 8.0.0 anyway_ and _No-one installing a new Moodle site on the minimum version of a dependency._ – I don't mind that there's resistance to changing this, but you keep assuming this could never really occur when it literally did a week ago.

scara:
Hi @leonstr,
for your consideration: the first GA is [`8.0.11`](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-11.html) while [`8.0.1`](https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-1.html) was still a dev one, though with that feature already available.

HTH,
Matteo

leonstr:
My point with this change is that the current text showing 8.0 as the minimum MySQL version implies that Moodle will work if one somehow has MySQL 8.0.0 whereas actually it does not work. And this specific issue has somehow arisen in the real world, so IMO consequently we should make this avoidable by including 8.0.1 as the minimum in the documentation and environment check. Why check the environment meets the requirements to run a Moodle version if we don't exclude a component version known not to work?

mickhawkins:
Hi all,

Thanks for the discussions and efforts thus far. Further to my comment on [MDL-85485](https://moodle.atlassian.net/browse/MDL-85485), I'm going to close this pull request, however I've added an alternative [pull request](https://github.com/moodle/devdocs/pull/1383) to add relevant information to the 4.5 MySQL requirements.

PR #1351, opened 2025-05-12
[docs] Provide full example for Forms API

fapdash:
- provide full example including `<?php` tag when providing an example file. If an example references a filename the code block should contain a complete example of this file.
- specify namespace according to the automatic class loading rules
- add backslash to `moodleform` for namespaced reference (`use moodleform` would also be fine)

---

Not sure if `use moodleform` is preferred to `extends \moodleform`. The [style guide](https://moodledev.io/general/development/policies/codingstyle#further-notes) seems to say that both variants are fine?


andrewnicols (approved):
Thanks for addressing this. I've backported these changes to stables.

PR #1354, opened 2025-05-16
Fix typo

davosmith:
There is a clear typo - the description indicates this is meant to read 'showDropZone' not 'hideDropZone' (which is documented immediately below).


ferranrecio (approved):
Thanks for submitting this fix. I will merge now.

PR #1360, opened 2025-05-23
Update moodleorg.md unlinked hidden courses

tallthom:
Korean, India, and Svenska


wildgirl (approved):
Thanks Thom for taking care of these updates.

PR #1373, opened 2025-06-12
[docs] Further clarify final deprecation procedure

ziegenberg:
If the method or function has a typed return, it needs to return accordingly, otherwise no return is required.

See [this comment](https://tracker.moodle.org/browse/MDL-78375?focusedId=1163095&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1163095) for reference.


ferranrecio (approved):
Seems a good change

PR #1376, opened 2025-06-17
[repo] Fix all tracker links to point to atlassian

andrewnicols:
Fixes ICT-5400


abgreeve (approved):
The changes here look correct to me. Andrew and I discussed some release links that are broken but will be fixed in a different issue.

PR #1377, opened 2025-06-17
[docs] Update list of issues in releases

andrewnicols:
ICT-5400 part 2


abgreeve (approved):
Thanks Andrew,
This brings us inline with our current way of displaying resolved issues for versions.

PR #1383, opened 2025-06-23
[docs] Add additional info to MySQL required version

mickhawkins:
This addresses the concerns of [MDL-85485](https://moodle.atlassian.net/browse/MDL-85485) without requiring a code change, by documenting the known reserved word issue with MySQL 8.0.0.

[MDL-85485]: https://moodle.atlassian.net/browse/MDL-85485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/general/releases/4.4.md
+++ b/general/releases/4.4.md
@@ -42,6 +42,14 @@ Since Moodle 4.3, the maximum length for the database prefix (`$CFG->prefix`) is
 
 :::
 
+:::danger Important
+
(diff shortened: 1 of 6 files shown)


andrewnicols (approved):
Personally I think this is just common sense. We really shouldn't need to document this kind of thing. Do NOT use development versions of things in production.

PR #1410, opened 2025-07-30
[MDL-80945] Update quizaccess page

michaelkotlyar:


data/main/components.json
```diff
@@ -34,6 +34,7 @@
         "qbank": "public/question/bank",
         "qbehaviour": "public/question/behaviour",
         "qformat": "public/question/format",
+        "quizaccess": "public/mod/quiz/access",
```
andrewnicols, line 37: Please remove this - it is wrong.

The new component is a subplugin of mod_quiz and not a global plugin type.
The content of this file is automatically pulled in.

docs/apis/plugintypes/quizaccess/_examples/override_rule.php
```diff
@@ -0,0 +1,48 @@
+use mod_quiz\local\access_override_rule_base;
+use MoodleQuickForm;
+use context_module;
+use context;
+
+class override_rule extends access_override_rule_base {
+
```
andrewnicols, line 7: Please pass this file through the code style checker and fix it.

It's worth noting tha twe also tend to further reduce line length limits to around 80 characters in docs because the way that code scrolls left-to-right makes it hard to view things in context.

docs/apis/plugintypes/quizaccess/_examples/override_rule.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * All of the below access_override_rule_base functions to be implemented.
+     */
+
+    public static function add_form_fields(context_module $context, int $overrideid, object $quiz, MoodleQuickForm $mform): void {
```
andrewnicols, line 12: Example of exagerated line length limit:
```suggestion
    public static function add_form_fields(
        context_module $context,
        int $overrideid,
        object $quiz,
        MoodleQuickForm $mform,
    ): void {
```

docs/apis/plugintypes/quizaccess/_examples/override_rule.php
```diff
@@ -0,0 +1,48 @@
...
+        // Do nothing.
+    }
+
+    public static function validate_form_fields(array $errors,
+            array $data, array $files, context_module $context): array {
```
andrewnicols, line 17: Note: This coding style is not accepted in Moodle core code any more.

versioned_docs/version-5.0/apis/plugintypes/quizaccess/_examples/override_rule.php
```diff
@@ -0,0 +1,49 @@
+use mod_quiz\form\edit_override_form;
+use mod_quiz\local\access_override_rule_base;
+use MoodleQuickForm;
+use context_module;
+use context;
+
+class override_rule extends access_override_rule_base {
```
andrewnicols, line 7: Ditto the other exampels

andrewnicols (changes requested):
Hi,

Firstly thanks for the really detailed update. Much appreciated.
Given this is documentation for a new feature, and we do not land new features in stable versions of Moodle (5.0), there is no need to document it for 5.0. I really appreciate you taking the time to do so, but it won't be landing there.

The only other comment I have is with regards to coding style.
For dev docs we artificially reduce the line length to ensure that code examples fit within the width of the code container. That's around about 105 characters max, but it's often easier to read when it's slightly less than that. I've included a screenshot of what I mean in the generated code:

Image

We also aim to make our coding examples meet current coding style rules because people will use them as copy/paste examples of good practice.

Thanks again.

PR #1413, opened 2025-08-08
5.0.2 minor release notes and 5.1 env update

mickhawkins:
5.0.2 minor release notes additions. I've also popped a second commit on to update the postgres min requirements on the 5.1 page to match the confirmed environment requirements.


general/releases/5.1.md
```diff
@@ -30,7 +30,7 @@ Moodle supports the following database servers. Again, version numbers are just
 
 | Database | Minimum version | Recommended |
 | --- | --- | --- |
-| [PostgreSQL](http://www.postgresql.org/) | 14 (increased since Moodle 4.5) | Latest |
+| [PostgreSQL](http://www.postgresql.org/) | 15 (increased since Moodle 5.0) | Latest |
```
andrewnicols, line 33: This was only increased in 5.1,not 5.0.

andrewnicols: Ah yes - that's it.  And yes, I think we do.

PR #1417, opened 2025-08-13
MIgrate phpunit docs

andrewnicols:
This is the start of migrating the phpunit docs.

We should land this and then add more detail re:
- use of attributes
- best practice
- `assertInstanceOf('some_stupid_string_class')` is bad


ferranrecio:
I reviewed the text and I think it is ready to pull.

PR #1427, opened 2025-08-25
MDL-86254: comment API

ferranrecio:
[docs] Comment API


cescobedo (commented):
It looks good to me, but there is a minor styling thing. On the Usage section, shouldn’t the lists of attributes end with a full stop at the end of each item? Apart from that, I will approve it if this is fixed.

PR #1440, opened 2025-09-29
[docs] Remove rule_overridable in quizaccess page

marxjohnson:
This feature was never implemented in this way, the docs are incorrect.


marxjohnson:
This resolves #1440

PR #1446, opened 2025-10-06
5.1 major release notes

mickhawkins:
I've also included one additional line in the minor releases, where an issue should have been included but was initially missed.

This includes updating the messaging about the directory restructure to match the latest links and information being published across various forums etc, for consistency.


HuongNV13:
Thanks, Mick
It looks good to me!

PR #1459, opened 2025-10-30
Fix PHP code example in enrol documentation

marcgeremias:


junpataleta (approved):

junpataleta:
Thanks, @marcgeremias. The fix looks correct, but it needs to be applied to the other equivalent versions of the affected page. I took the liberty of doing this on your behalf. 

Cheers.

PR #1486, opened 2025-12-05
Fix get_records_list() default param values

leonstr:


paulholden (changes requested):
Thanks Leon, the `$sort` parameter is also wrong, let's fix that too

PR #1488, opened 2025-12-09
Update coding standard regarding function definitions using $options arrays as parameters

traviswindsormoodle:
As per MDL-80349, updating the devdocs appropriately

### DIFF
--- a/general/development/policies/codingstyle/index.md
+++ b/general/development/policies/codingstyle/index.md
@@ -1159,6 +1159,63 @@ Magic methods are heavily discouraged, justification will be required when used.
 
 (See [MDL-52634](https://moodle.atlassian.net/browse/MDL-52634) for discussion of rationale)
 
+### Using arrays for options as arguments
+
+All arguments to a function should be explicitly listed out and defined with a type. Using arrays like this can result in errors as the types are not enforced as well as leading to poor documentation of the function.
+
+<InvalidExample>
+
+```php
+public function bad_function(
+    string $text,
+    ?context $context = null,
+    array $options = [],
+): string;
+```
+
+</InvalidExample>
+
+Instead of a generic array of options, each option should be specified as a function parameter. Optional arguments should provide appropriate defaults.
+
+<ValidExample>
+
+```php
+function goodfunction(
+    string $text,
+    ?context $context = null,
+    bool $trusted = false,
+    ?bool $clean = null,
+    bool $filter = true,
+    bool $para = true,
+    bool $newlines = true,
+    bool $overflowdiv = false,
+    bool $blanktarget = false,
+    bool $allowid = false,
+): string;
+```
+
+</ValidExample>
+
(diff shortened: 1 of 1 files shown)


general/development/policies/codingstyle/index.md
```diff
@@ -1159,6 +1159,45 @@ Magic methods are heavily discouraged, justification will be required when used.
...
+public function badfunction(
+    string $text,
+    ?context $context = null,
+    array $options = [],
+): string;
+```
+
+</InvalidExample>
+
+If arguments are optional, then they should be marked appropriately as optional with a default value.
```
andrewnicols: This is just standard PHP. Not sure if we need to mention this.

I'd suggest the more pertinent point is that individual options should be explicitly specified:

```suggestion
Instead of a generic array of options, each option should be specified as a function parameter. Optional arguments should provide appropriate defaults.
```

general/development/policies/codingstyle/index.md
```diff
@@ -1159,6 +1159,45 @@ Magic methods are heavily discouraged, justification will be required when used.
...
+    bool $newlines = true,
+    bool $overflowdiv = false,
+    bool $blanktarget = false,
+    bool $allowid = false,
+): string;
+```
+
+</ValidExample>
+
+If there was a very good reason, then the use of an array could be considered. But it is strongly encouraged that all parameters be explicit and typed.
```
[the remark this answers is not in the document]
traviswindsormoodle: The Jira ticket doesn't have explicit examples about exceptions, it just acknowledges them. The devdocs should also be complete enough to stand on their own without references to Jira tickets in lieu of details

general/development/policies/codingstyle/index.md
```diff
@@ -1159,6 +1159,45 @@ Magic methods are heavily discouraged, justification will be required when used.
...
+    bool $para = true,
+    bool $newlines = true,
+    bool $overflowdiv = false,
+    bool $blanktarget = false,
+    bool $allowid = false,
+): string;
+```
+
+</ValidExample>
+
```
andrewnicols: We should also mention the alternative, which is to use a dedicated options class.


```suggestion

Alternatively a dedicated options class may be used - for example:

<ValidExample>

```php
public function good_function(
    string $text,
    ?context $context = null,
    filter_options $options = null,
): string;

</ValidExample>

```

```

general/development/policies/codingstyle/index.md
traviswindsormoodle: I don't mind your suggestion. It was more that a key motivation for using an array is to handle optional parameters "efficiently" and I was addressing the motivation

general/development/policies/codingstyle/index.md
traviswindsormoodle: I didn't explicitly include this as it felt redundant. It's just a explicitly typed parameter, it's just that the type is custom class.

PR #1495, opened 2026-01-06
Note about sunset of Telegram

andrewnicols:


wildgirl (approved):
The sunset of Telegram is clear. Thanks Andrew.

PR #1504, opened 2026-01-19
[docs] Add class properties deprecation guidelines

meirzamoodle:
Documents the process for deprecating class properties, which cannot emit runtime warnings in PHP < 8.4.

Key points:
- Private properties: remove immediately
- Protected properties in non-extensible classes: can be removed
- Public/protected properties in extensible classes: use @deprecated tags, attributes, and provide getter/setter methods with warnings
- From Moodle 6.0 (PHP 8.4+): use property hooks for proper deprecation


meirzamoodle:
Hi, @andrewnicols 

The document covers everything you mentioned a couple of days ago. I added an example for setter and getter. Please review this PR.

Thanks.

PR #1513, opened 2026-02-03
Migrate Git commit cheat sheet

junpataleta:
Migrate the Git commit cheat sheet documentation to the dev docs.

Also, some improvements:

- Remove the following: 
  - Testing instructions bit - It's already covered in `/general/development/process/testing`
  - Removing strings - Also already covered in the String deprecation docs.
- Move the `Introducing new strings` section under the `Language`  section of the Coding style docs.

### DIFF
--- a/data/migratedPages.yml
+++ b/data/migratedPages.yml
@@ -106,6 +106,9 @@ Coding:
 Coding_style:
 - filePath: "/general/development/policies/codingstyle/index.md"
   slug: "/general/development/policies/codingstyle"
+Commit_cheat_sheet:
+- filePath: "/docs/guides/git/cheatsheet.md"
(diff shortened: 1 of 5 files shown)


andrewnicols (approved):

ziegenberg:
> This looks good to me. The only thing that gives me cause for concern is whether this belongs in versioned docs or whether it would be better in somewhere like:
> 
> https://moodledev.io/general/development/abc
> 
> Any thoughts @junpataleta ?

I second @andrewnicols concern and suggest putting it under https://moodledev.io/general/development/policies/codingstyle/commit-cheatsheet. Also, I suggest renaming the file to `commit-cheatsheet.md` as it's more specific.

PR #1562, opened 2026-04-13
[docs] Document the JS Deprecation utilities and process

andrewnicols:

PR #1566, opened 2026-04-14
Document template documentation

andrewnicols:
This is based on #1562 and also includes a couple of more general deprecation notes.


safatshahin (approved):
Thanks @andrewnicols for updating this doc, looking good.

PR #1582, opened 2026-04-29
[config] Enable sitemap lastmod and align with Docusaurus v4 defaults

### LINKED ISSUE #1579: [chore] Add “lastmod” to sitemap generation
### What do we need to do?

Lately I’ve been working with some sitemap-enabled crawlers and one of the techniques used often to avoid having to crawl every single page/etag is as simple as adding the “lastmod” information to every sitemap entry.

And it seems that Docusaurus supports it pretty well since version 3.2:

https://docusaurus.io/blog/releases/3.2#sitemap-lastmod

The datetime defaults to the datetime of the last commit of the file, not the build date. And that’s perfect.

I’d recommend to add it ASAP, it’s just a setting in config. And I’m sure that will help, in fact I’m right now building one of those sitemap-based crawlers for a project…and can test it immediately.

Ciao :-)

stronk7:
Configure the sitemap plugin (via classic preset) with lastmod: 'date', priority: null and changefreq: null, matching the upcoming v4.0 defaults as described in https://docusaurus.io/blog/releases/3.2#sitemap-lastmod.

Fixes #1579

I've tested it locally and is working well, all the URLs based on git files are getting the `<lastmod>YYYY-MM-13DD</lastmod>` element.

Obviously, those generated from JSON or other methods, not coming from git (tags, search page, ...) do not have that new information.

Also, at the same time, the old (and useless) `priority` and `changefreq` have been removed, aligning which the expected defaults for v4.0.

Ciao :-)

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -115,6 +115,11 @@ const config = {
                         trackingID: 'G-L9EE8RW5B1',
                     }
                     : undefined,
+                sitemap: {
+                    lastmod: 'date',
+                    priority: null,
+                    changefreq: null,
+                },
             }),
         ],
     ],


stronk7:
Just noting that it's also possible to use `datetime` instead of `date` if we want more resolution, but I don't think that's needed.

In any case I hope this helps to reduce full crawling, because crawlers will have less work to do.

Ciao :-)

andrewnicols (approved):

PR #1602, opened 2026-05-29
[docs] codingstyle: Fix SQL indentation image alt text

### LINKED ISSUE #1395: Accessibility - SQL Indentation image alt text is the file name.
https://github.com/moodle/devdocs/blob/8a7b2f9c339fc6227f0016a8cd06b5c763dd2679/general/development/policies/codingstyle/sql.md?plain=1#L51

Produces an `img` element with `alt="sql_indentation.png"`. Instead of providing the full description in the alternative text, I encourage providing a visible image caption so that everyone can benefit from a text description of the image's purpose.

travis-kuenzi:
## Summary

Fixes #1395.

This PR updates the SQL coding style documentation by replacing the filename-based alt text for the SQL indentation image with a meaningful description. It also adds a visible text description below the image so the purpose of the image is available to all readers.

## Testing

- Previewed the Markdown change in GitHub.
- Confirmed the image reference still points to `./_sql/sql_indentation.png`.
- Confirmed the image alt text no longer uses the file name.
- Confirmed a visible description appears below the image.


jrchamp:
I love this effort you've made. I do want to note that now that the text is duplicated, it will be read by a screen reader twice. "Visible" captions are my preference because they can highlight the purpose of the image in a way that might otherwise need to be inferred. To that end, I also encourage another word or two to clarify that the "vertical alignment" means uses spaces to horizontally align the last letter of these SQL keywords into the same vertical column. If that is already redundant with the surrounding text, then the image can delightfully become a redundant, functionally decorative, visual example of clear text-based documentation.

PR #1604, opened 2026-06-02
[docs] Update route_testcase namespace

andrewnicols:
Pinging @jrchamp


jrchamp:
Copilot's suggestions are all the same wording change: switch from "and having routing it" to "and routing it". My guess is that "and having it route" is the other wording option.

PR #1615, opened 2026-06-13
docs: clarify capability contextlevel guidance

### LINKED ISSUE #1339: [docs] Improve documentation around the contextlevel in the docs

https://moodledev.io/docs/5.0/apis/subsystems/access#how-to-define-new-capabilities-in-plugins

In the dev chat there was a discussion around what `contextlevel` on capabilities does and means we seemed to come up with a good description that is not included in the docs at the moment:

> It indicates what context you intend to check the capability at, and does control where it appears, for instance moodle/site:config is defined as CONTEXT_SYSTEM and that means you can't change it at lower levels. If you go to a course (or category, or module or block, or user for that matter) and try to override capabilities there, that one doesn't show on the list.  Because if you did override it in a course, it wouldn't make any difference when you check it as context system.

It seems that adding this or something similar will help people understand what `contextlevel` they should choose to use better.

None

ded-furby:
## Summary
- explain that `contextlevel` should match the highest context where a capability is checked and overridden
- add the override-screen example for `moodle/site:config`
- backport the clarification to the current docs plus versions 4.5, 5.0, 5.1, and 5.2

Closes #1339

## Validation
- `corepack yarn mdlint docs/apis/subsystems/access.md versioned_docs/version-4.5/apis/subsystems/access.md versioned_docs/version-5.0/apis/subsystems/access.md versioned_docs/version-5.1/apis/subsystems/access.md versioned_docs/version-5.2/apis/subsystems/access.md`
- repository `lint-staged` hook (`markdownlint-cli2 --fix` and `cspell --no-must-find-files --no-progress`) ran during commit


docs/apis/subsystems/access.md
```diff
@@ -66,6 +66,8 @@ Where the meaning of array keys is:
...
 | `clonepermissionsfrom` | when you are adding a new capability, you can tell Moodle to copy the permissions for each role from the current settings for another capability. This may give better defaults than just using archetypes for administrators who have heavily customised their roles configuration. The full syntax is: `clonepermissionsfrom` => `moodle/quiz:attempt` |
 
+When choosing `contextlevel`, pick the highest context where the capability is expected to be checked and overridden. Moodle can still check the capability in lower contexts, but it only appears in override screens where changing it can affect the result. For example, `moodle/site:config` uses `CONTEXT_SYSTEM`, so it is not shown in course or activity overrides because changing it there would have no effect when the capability is checked at system context.
```
jrchamp, line 69: By highest, do you mean "largest number"? Initially, I thought of "highest point, hierarchically", where CONTEXT_SYSTEM is the top of the pyramid and each "lower" level has more items (and thus more width).

Here is an idea that I had (it's not perfect, but I hope it is helpful):

When choosing `contextlevel`, use the most specific level where the capability can be overridden. A permission can only be managed at the chosen level or levels with a more general scope. For example, `moodle/site:config` uses `CONTEXT_SYSTEM`, so it can only be set or overridden at the system context (the widest scope). Moodle intentionally hides the capability from override screens that have a more specific scope.

PR #1626, opened 2026-06-29
MDLSITE-5960 [docs] Ban use of resetAfterTest in shared setup

safatshahin:
Included stronger recommendation and red flags for this matter.


ferranrecio (approved):
The explanation is clear and goes to the point.

PR #1638, opened 2026-07-16
[docs] Add security announcements to 5.2.1 and friends

mickhawkins:


raortegar (approved):
Thanks, Mick.
	It looks good to me.

PR #1640, opened 2026-07-19
Improve dependency injection guidance

cameron1729:
## Background
Existing DI documentation focuses heavily on accessing Moodle's DI container, which can blur the distinction between constructor injection and using the container as a service locator.

This distinction is becoming increasingly important as the core container sees wider adoption. Moodle needs clear guidance showing that classes should declare only their immediate dependencies, code to contracts, and remain unaware of the container. Container lookups should be restricted to composition roots or compatibility boundaries.

The Calendar API is a useful canonical example. It has used constructor injection since Moodle 3.3, before Moodle had a native DI container, and already contains a substantial graph of factories, mappers, policies, retrieval strategies, and output services. There is [work in progress to migrate it to native DI](https://moodle.atlassian.net/browse/MDL-89216) - and the documentation here references that work.

## Changes
This PR:

  - Defines dependency injection separately from the DI container
  - Recommends constructor injection as the normal approach
  - Explains application boundaries and composition roots, including framework owned composition through routing
  - Distinguishes runtime values from container-managed services
  - Clarifies what autowiring can infer and when a container definition is required
  - Explains why injecting the container creates a service locator
  - Documents the related [Moodle and Moodle Extra PHP CodeSniffer rules](https://github.com/moodlehq/moodle-cs/pull/228)
  - Uses the Calendar API as a substantial core example of composing a dependency graph from a small number of boundary lookups (**NB**: the examples used are based on ongoing work in [MDL-89216](https://moodle.atlassian.net/browse/MDL-89216))
  - Updates the Clock API guidance to follow the same constructor-injection model.

## Related work
  - Calendar API migration to native DI container: [MDL-89216](https://moodle.atlassian.net/browse/MDL-89216)
  - Coding standard guidance: [moodlehq/moodle-cs#228](https://github.com/moodlehq/moodle-cs/pull/228)

> [!IMPORTANT]
> This PR documents Calendar APIs introduced by MDL-89216 and coding-standard rules introduced by moodlehq/moodle-cs#228. It must not be merged before those have landed.
>
> Moodle `main` currently targets Moodle 5.3, scheduled for release on 5 October 2026. If MDL-89216 is integrated for Moodle 5.3, this PR should merge after its integration and before the 5.3 developer documentation is forked. If the issue moves to a later release, this PR should be held until the developer documentation targets the release which contains it.

[MDL-89216]: https://moodle.atlassian.net/browse/MDL-89216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/apis/core/clock/index.md
+++ b/docs/apis/core/clock/index.md
@@ -11,7 +11,7 @@ description: Fetching the current time
 
 <Since version="4.4" issueNumber="MDL-80838" />
 
-Moodle supports use of a [PSR-20](https://php-fig.org/psr/psr-20/) compatible Clock interface, which should be accessed using Dependency Injection.
+Moodle provides a [PSR-20](https://php-fig.org/psr/psr-20/) compatible Clock interface. Classes which need the current time should receive `\core\clock` through constructor injection.
 
 This should be used instead of `time()` to fetch the current time. This allows unit tests to mock time and therefore to test a variety of cases such as events happening at the same time, or setting an explicit time.
 
@@ -23,34 +23,21 @@ We recommend that the Clock Interface is used consistently in your code instead 
 
 ## Usage {/* #usage */}
 
-The usage of the Clock extends the PSR-20 Clock Interface and adds a new convenience method, `\core\clock::time(): int`, to simplify replacement of the global `time()` method.
+The Moodle Clock extends the PSR-20 Clock Interface and adds the convenience method `\core\clock::time(): int` to simplify replacement of the global `time()` function.
 
-### Usage in standard classes {/* #usage-in-standard-classes */}
+### Usage via constructor injection {/* #usage-via-constructor-injection */}
 
-Where the calling code is not instantiated via Dependency Injection itself, the simplest way to fetch the clock is using `\core\di::get(\core\clock::class)`, for example:
+Declare the clock as a constructor dependency:
 
-```php title="Usage in legacy code"
-$clock = \core\di::get(\core\clock::class);
-
-// Fetch the current time as a \DateTimeImmutable.
-$clock->now();
-
-// Fetch the current time as a Unix Time Stamp.
-$clock->time();
-```
-
-### Usage via Constructor Injection {/* #usage-via-constructor-injection */}
-
-The recommended approach is to have the Dependency Injector inject into the constructor of a class.
-
-```php title="Usage in injected classes"
+```php title="Using an injected clock"
 namespace mod_example;
 
 class post {
     public function __construct(
         protected readonly \core\clock $clock,
         protected readonly \moodle_database $db,
-    )
+    ) {
+    }
 
     public function create_thing(\stdClass $data): \stdClass {
         $data->timecreated = $this->clock->time();
@@ -62,17 +49,35 @@ class post {
 }
 ```
 
-When using DI to fetch the class, the dependencies will automatically added to the constructor arguments:
+At the application boundary, obtain the highest-level service. The container supplies its clock and database dependencies:
 
 ```php title="Obtaining the injected class"
-$post = \core\di::get(post::class);
+$post = \core\di::get(\mod_example\post::class);
 ```
+
+Do not call `\core\di::get(\core\clock::class)` from inside `post`. That hides the dependency and uses the container as a service locator.
+
+### Legacy and procedural boundaries {/* #usage-in-standard-classes */}
+
+Code which cannot receive constructor dependencies without a backwards-incompatible change may fetch the clock at its procedural or static boundary:
+
+```php title="Compatibility usage in legacy code"
+$clock = \core\di::get(\core\clock::class);
+
+// Fetch the current time as a \DateTimeImmutable.
+$clock->now();
+
+// Fetch the current time as a Unix timestamp.
+$clock->time();
+```
+
+Keep this lookup at the boundary. Pass the clock into any objects created below it.
 
 ## Unit testing {/* #unit-testing */}
 
-One of the most useful benefits to making consistent use of the Clock interface is to mock data within unit tests.
+One of the most useful benefits of consistently using the Clock interface is the ability to control time in unit tests.
 
-When testing code which makes use of the Clock interface, you can replace the standard system clock implementation with a testing clock which suits your needs.
+Calling either `advanced_testcase` helper described below performs the complete replacement: it creates a test clock, calls `\core\di::set(\core\clock::class, $clock)` to replace the container's clock for the test, and returns that same object. No additional container configuration is required. Any container-managed service resolved afterwards receives the replacement clock through constructor injection. This is an example of [replacing a dependency before obtaining the aggregate root](../di/index.md#unit-testing).
 
 :::tip[Container Reset]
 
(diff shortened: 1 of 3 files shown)


timhunt:
Thanks for writing this @cameron1729, it looks like important advice about best practice (about which I certainly know less than I should, and I just learned things.)

I am going to add a few comments about suggested improvements, but I will leave a full review to someone more knowledgeable.

docs/apis/core/di/index.md
```diff
@@ -5,225 +5,458 @@ tags:
...
+
+## Constructor injection {/* #constructor-injection */}
+
+Constructor injection is one form of dependency injection. Dependencies can also be supplied through methods, including setters, or through properties. Moodle code should normally use constructor injection for required collaborators so that every instance is complete and its dependencies remain explicit.
+
+Constructor injection is not a feature of Moodle's DI container. For required collaborators, it is a long-established PHP design practice that has been widely used for well over a decade. A class declares only its immediate dependencies and remains unaware of whether they were supplied by application code, a test, hand-written composition code, or a container. This makes the class easier to understand, reuse, and test. It also makes the class naturally compatible with Moodle's container, which can use its constructor types to assemble the object graph. The [unit testing example](#testing-a-class-without-a-container) shows the same class being constructed directly with test doubles.
```
timhunt, line 41: https://xkcd.com/285/ - for people who want to know more, and to strengthen your argument that this is "a long-established PHP design practice", it would be good to link to an in-depth discussion of DI somewhere on the web, if you know a good one.

docs/apis/core/di/index.md
```diff
@@ -5,225 +5,458 @@ tags:
...
     ) {
     }
+
+    public function create_for_user(int $userid): void {
+        $this->repository->create(
+            userid: $userid,
+            timecreated: $this->clock->time(),
+        );
+    }
 }
```
timhunt, line 69: Given that a good example is worth a thousand words, I am going to say, I did not find this the most intuitive example. It took me some time to pick it apart, and I might not have been motivated to do that.

I can see you have tried to make the example quite generic (mod_example, local, service, .... It might actually be better to go more specific, to make the purpose of things clearer. I can't tell what sort of thing create_for_user is creating. It does not even return the thing created.

So, if there are any ways you can make this example more self-explanatory, or self-motivating, I think it might be worth putting some time into that.

docs/apis/core/di/index.md
```diff
@@ -5,225 +5,458 @@ tags:
...
-$thing = \core\di::get(my_thing::class);
+With the interface mapping configured, Moodle's container can construct `reminder_service` and recursively supply its dependencies when application code requests the top-level service through `\core\di::get(\mod_example\local\reminder_service::class)`. The next section explains where such top-level requests belong.
+
+## Boundaries and composition roots {/* #composition-root */}
+
+Where `\core\di::get()` is called must be considered carefully. Moodle spans more than two decades of architectural styles, so its entry points are not all composed in the same way.
+
+Start by identifying the operation's application boundary: the outermost component code invoked for that operation. Common boundaries include route controller methods, external functions, hook callbacks, scheduled task `execute()` methods, and CLI or legacy PHP scripts. Runtime values such as validated request parameters and the current user ID enter at the boundary. It then invokes the top-level application objects for that operation. The classes called beneath it are inside the application's object graph; they are not boundaries merely because they call other classes.
```
timhunt, line 84: I would not call them 'legacy PHP scripts' yet. As you say, this is how Moodle has worked for 20+ years. They are currently the main entry point into most of Moodle, so I would list them first, perhaps ("top level PHP scripts like `course/view.php`).

cameron1729:
Thank you very much for taking the time to read this @timhunt - your feedback as someone wanting to learn more about DI and best practice is extremely valuable. This is something I'm working on in my personal time, as well as an accompanying example in core which I hope can serve as a kind of reference implementation. So I might be slow to get back to your points/comments but I will definitely be using your suggestions to improve this document :)

Thanks again.

timhunt:
Thanks @cameron1729. It might be best to get this first version published, and open a new issue for further improvements later.

And, when this is published, I wonder if you can get a shout-out in integration exposed? Having the docs is one thing. Even better if people actually read them! :-)

docs/apis/core/clock/index.md
```diff
@@ -11,7 +11,7 @@ description: Fetching the current time
 
 <Since version="4.4" issueNumber="MDL-80838" />
 
-Moodle supports use of a [PSR-20](https://php-fig.org/psr/psr-20/) compatible Clock interface, which should be accessed using Dependency Injection.
+Moodle provides a [PSR-20](https://php-fig.org/psr/psr-20/) compatible Clock interface. Classes which need the current time should receive `\core\clock` through constructor injection.
```
andrewnicols: This is not universally true. Constructor injection is the preferred approach, but is not available in all cases - primarily not in legacy cases such as outside of a class, in global functions, and so on. We currently use DI as a B/C approach to support migration so that when we move a function into an instance method we can polyfill it:

```php
function exists_auth_plugin($auth) {
    return \core\di::get(\core\authentication::class)->plugin_exists($auth);
}
```

We are also forced to use it in classes which are not currently loaded using DI, though we could instead look at using `\core\di::get_container()->injectOn($this);` in the constructor for these cases (though we'd like add a static wrapper to simplify this).

It's also not true for static methods since they have no constructor. I'm not sure of the best recommendation in these cases.

I think it's fine to keep the line you have now, but perhaps to add a reference to the DI documentation here and/or add a note about fetching using other DI approaches for cases where constructor injection is not available. Or just keep the existing language since constructor injection is a form of DI.

docs/apis/core/clock/index.md
```diff
@@ -62,17 +49,35 @@ class post {
...
+
+Keep this lookup at the boundary. Pass the clock into any objects created below it.
+
 ## Unit testing {/* #unit-testing */}
 
-One of the most useful benefits to making consistent use of the Clock interface is to mock data within unit tests.
+One of the most useful benefits of consistently using the Clock interface is the ability to control time in unit tests.
 
-When testing code which makes use of the Clock interface, you can replace the standard system clock implementation with a testing clock which suits your needs.
+Calling either `advanced_testcase` helper described below performs the complete replacement: it creates a test clock, calls `\core\di::set(\core\clock::class, $clock)` to replace the container's clock for the test, and returns that same object. No additional container configuration is required. Any container-managed service resolved afterwards receives the replacement clock through constructor injection. This is an example of [replacing a dependency before obtaining the aggregate root](../di/index.md#unit-testing).
```
andrewnicols: Probably also worth noting that it is automatically reset between tests so there is no additional cleanup required.

docs/apis/core/di/index.md
```diff
@@ -5,225 +5,458 @@ tags:
   - Container
   - PSR-11
   - PSR
-description: The use of PSR-11 compatible Dependency Injection in Moodle
+description: Using constructor injection and the Moodle dependency injection container
```
andrewnicols: Disagree with this change. This documentation covers the use of DI, not the use of constructor injection. CI is just one use of DI, not the only.

PR #1649, opened 2026-07-23
[docs] Initial draft of the REST style guide

andrewnicols:
This is the initial draft of the REST style guide for RFC.

### DIFF
--- a/docs/guides/rest/style.md
+++ b/docs/guides/rest/style.md
@@ -1,0 +1,1030 @@
+---
+title: Style Guide
+tags:
+  - rest
+  - api
+  - style
+description: Standards and conventions for designing modern REST APIs in Moodle.
+---
+
+## 1. Purpose
+
+This document defines the core architectural, semantic, and structural design principles for Moodle's modern REST API layer. It serves as the single source of truth for REST semantics, guiding both human developers and AI engines in creating predictable, scalable, and idiomatic integrations.
+
+### Objectives
+
+- standardize REST API behaviour
+- improve API consistency
+- improve OpenAPI generation quality
+- align APIs with Moodle architecture
+- support long-term maintainability
+- support deterministic AI-assisted specification generation
+
+### Scope
+
+This document defines:
+
+- REST resource design
+- HTTP semantics
+- naming conventions
+- URI conventions
+- response conventions
+- error conventions
+- versioning behaviour
+- Moodle-specific REST constraints
+
+This document does NOT define:
+
+- implementation details
+- database design
+- business logic
+- frontend behaviour
+
+---
+
+## 2. Rule Types
+
+The style guide separates:
+
+| Section | Purpose |
+| --- | --- |
+| Normative Rules  | mandatory behaviour  |
+| Canonical Patterns  | reusable API patterns  |
+| Rationale  | explanatory guidance  |
+
+Normative sections MUST use only:
+
+- MUST
+- MUST NOT
+- REQUIRED
+- FORBIDDEN
+- SHOULD
+- SHOULD NOT
+- MAY
+
+---
+
+## 3. Core REST Principles
+
+All Moodle APIs MUST satisfy **Richardson Maturity Model Level 2**.
+
+### 3.1 Resource-Oriented Design
+
+API URIs MUST represent resources (nouns), never actions or operations (verbs).
+
+Standard CRUD processing logic MUST be expressed strictly through HTTP semantics (GET, POST, PUT, PATCH, DELETE), NOT by appending action triggers to the path string.
+
+#### REQUIRED
+
+Resource-oriented APIs:
+
+```http
+GET /courses
+GET /courses/{courseid}
+POST /courses
+PATCH /courses/{courseid}
+DELETE /courses/{courseid}
+GET /course-categories
+GET /users
+```
+
+#### FORBIDDEN
+
+RPC-style APIs:
+
+```http
+POST /core_course_get_courses
+POST /createCourse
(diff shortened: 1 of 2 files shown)


docs/guides/rest/style.md
```diff
@@ -0,0 +1,1030 @@
...
+
+The style guide separates:
+
+| Section | Purpose |
+| --- | --- |
+| Normative Rules  | mandatory behaviour  |
+| Canonical Patterns  | reusable API patterns  |
+| Rationale  | explanatory guidance  |
+
+Normative sections MUST use only:
```
ziegenberg, line 55: Can we quote or reference RFC 2119 and 8174 here? I know, this is not an Internet Standard, but RFC 2119 was created for a reason, and IMHO it fits very well here, too.

```suggestion
Normative sections MUST use only the following key words, and they are to be interpreted as described in [BCP 14](https://datatracker.ietf.org/doc/html/bcp14) [[RFC2119](https://datatracker.ietf.org/doc/html/rfc2119)] [[RFC8174](https://datatracker.ietf.org/doc/html/rfc8174)] when, and only when, they appear in all capitals:
```

docs/guides/rest/style.md
```diff
@@ -0,0 +1,1030 @@
...
+- FORBIDDEN
+- SHOULD
+- SHOULD NOT
+- MAY
+
+---
+
+## 3. Core REST Principles
+
+All Moodle APIs MUST satisfy **Richardson Maturity Model Level 2**.
```
ziegenberg, line 69: We should include a link to a normative source so that people unfamiliar with the model can read about it.

```suggestion
All Moodle APIs MUST satisfy **[Richardson Maturity Model Level 2](https://martinfowler.com/articles/richardsonMaturityModel.html)**.
```

docs/guides/rest/style.md
```diff
@@ -0,0 +1,1030 @@
...
+| 200  | Success  |
+| 201  | Created  |
+| 202  | Accepted  |
+| 204  | No Content  |
+| 400  | Invalid request  |
+| 401  | Unauthenticated  |
+| 403  | Unauthorized  |
+| 404  | Resource not found  |
+| 409  | Conflict  |
+| 422  | Validation failure  |
```
ziegenberg, line 884: Section 15.1 also mentions `412 Precondition Failed`, so we should include it here.

```suggestion
| 404  | Resource not found  |
| 409  | Conflict  |
| 412  | Precondition Failed  |
| 422  | Validation failure  |
```

docs/guides/rest/style.md
```diff
@@ -0,0 +1,1030 @@
...
+
+### 15.2 Idempotency
+
+| Method | Idempotent |
+| --- | --- |
+| GET  | Yes  |
+| PUT  | Yes  |
+| DELETE  | Yes  |
+| POST  | No  |
+| PATCH  | No guarantee  |
```
jtse, line 967: Minor: Can we simplify PATCH just "No" instead of "No guarantee"?

docs/guides/rest/style.md
```diff
@@ -0,0 +1,1030 @@
...
+
+Retryable POST operations SHOULD support:
```
jtse, line 969: Thoughts about upgrading Idempotency-Key from SHOULD to MUST?

1. This would make retry safety a first principle. See my other comment below.
2. In general, fewer optionality reduces ambiguity and thus complexity.

Of course, the downside is slightly more client-side complexity. But in my opinion is well worth not having to deal with support calls about duplicate records due to retries.

docs/guides/rest/style.md
```diff
@@ -0,0 +1,1030 @@
...
+
+```http
+Idempotency-Key
+```
+
+Duplicate request behaviour MUST be explicitly defined.
+
+#### RATIONALE
+
+Clients on unreliable networks, particularly mobile, need to safely retry requests without accidentally creating duplicate resources; since POST is inherently non-idempotent, an explicit Idempotency-Key lets the server recognize and deduplicate a retried creation request as the same logical operation, rather than the client having to guess whether a prior request actually succeeded before retrying.
```
jtse, line 979: It's not just network unreliability but rather unreliability in general, which would include server overload, especially during the start of semester (leading to duplicate requests).

jtse (commented):
Thanks for writing this! A general point: Consider having fewer SHOULDs, either upgrading them to MUST or removing them. SHOULD in specs tend to make things more complicated.

docs/guides/rest/style.md
```diff
@@ -0,0 +1,1030 @@
...
+## 7. Resource Structure Rules
+
+### 7.1 Nested Resources
+
+Nested resources MUST only be used only for direct ownership or containment relationships.
+
+Rules:
+
+- must represent clear hierarchical ownership
+- must remain shallow (maximum recommended depth: 2)
```
jtse, line 325: +1 to shallow resources (especially with respect to the technical reasons you outlined below). Where would a common resource like student submissions for an activity (assignment) go? If it's nested under activities it would be a depth of 3.

On the other hand, that nesting also feels intuitive to me as a dev. Maybe one way to resolve this is the top level resource courses (and enrollment) don't change often especially later in the school year/semester, so it's highly cacheable.

Thoughts?

jtse (commented):
Looking forward to seeing how this evolves and prototype implementations.

docs/guides/rest/style.md
andrewnicols: Thanks - I had a note to link to this but missed it when moving from Confluence to Markdown /o\

PR #1671, opened 2026-08-11
Workplace version data for August '26 minors.

paulholden:
Correct 5.1R support date (now EOL).


sarjona (approved):
Thanks @paulholden for creating this patch to update the version data.

It looks good :-)

PR #1691, opened 2026-09-11
[docs] Document API scopes

andrewnicols:


docs/apis/subsystems/routing/scopes.md
```diff
@@ -0,0 +1,185 @@
...
+
+A route with **no** `#[scopeset]` attribute at all currently behaves identically, at runtime, to one marked `#[unscoped_resource]`: neither requires any scope. However, this is purely incidental, and should not be relied upon -- the absence of scopes must always be a deliberate, reviewable decision, expressed with `#[unscoped_resource]`, rather than an oversight.
+
+When adding tests for your own routes, use the `assert_route_is_scoped()` or `assert_route_is_unscoped()` helpers described below.
+
+:::
+
+## How scope validation works {/* #how-scope-validation-works */}
+
+Scope requirements are read from the current route by the `\core\router\middleware\moodle_oauth2_scope_attribute_middleware`, which attaches the resolved `scopeset[]` to the request as an attribute.
```
meirzamoodle: No `moodle_oauth2_scope_attribute_middleware` exists anywhere in the branch.

The actual class is: `public/lib/classes/router/middleware/moodle_scope_attribute_middleware.php`
-> `\core\router\middleware\moodle_scope_attribute_middleware`

docs/apis/subsystems/routing/defining-scopes.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+- The **summary** is a short, human-readable label for the scope, for example as shown in a list of scopes available to grant to a client.
+- The **description** is a longer explanation of what the scope allows, for example as shown on a consent screen, or in the generated OpenAPI documentation.
+
+Both attributes extend `\core\lang_string`, so they take the same arguments as `get_string()` -- a string identifier and a component, with an optional `$a` object -- and are lazily resolved to the current language only when their `get_summary()` / `get_description()` accessor is called.
+
+Core's convention is to key these lang strings after the scope's identifier (without the component prefix), for example:
+
+```php title="course/lang/en/course.php"
```
meirzamoodle: Shouldn't be `lang/en/course.php` ?

docs/apis/subsystems/routing/scopes.md
```diff
@@ -0,0 +1,185 @@
...
+
+<Since issueNumber="MDL-89710" version="5.3" />
+
+API routes are protected by Moodle's normal login and capability checks, but when a request is authenticated with an OAuth2 access token, or a Personal Access Token, the router also enforces **Scopes**. A scope is a small, named permission (for example `core_user:user:read`) that describes what a particular token is allowed to do.
+
+:::info[Why do we need scopes?]
+
+The primary purpose of scopes is to limit the ability of a particular client or integration rather than giving the integration the same abilities as the owner.
+
+For example if an administrator wishes to set up an integration with a tool which updates their user profile whenever they complete a run, then they would want to limit the integration to only performing user read and write actions and not making any other system configuration changes.
```
meirzamoodle: missing comma after "For example"

docs/apis/subsystems/routing/defining-scopes.md
```diff
@@ -0,0 +1,147 @@
...
+`abstract_scope::get_identifier()` builds the full, colon-separated scope identifier by walking up the class hierarchy from the leaf class, collecting the `#[identifier_attribute]` value declared at each level, until it reaches `\core\router\scope\abstract_scope` itself. It then prepends the Frankenstyle component name, taken from the leaf class's own namespace. For the `read` class above, this produces:
+
+```text
+core_course : course : content : read
+```
+
+that is `core_course:course:content:read`.
+
+If any class in the hierarchy is missing its `#[identifier_attribute]`, a `coding_exception` is thrown when the identifier is resolved.
+
```
meirzamoodle, line 94: I think we need a note here. The `coding_exception` is only surfaced when the identifier is resolved directly. Dduring scope discovery it's caught and downgraded to a debugging() message, and the class is skipped:

```
// public/lib/classes/oauth2/server/repository/scope_repository.php:137
} catch (\Throwable $e) {
    debugging("Skipping scope class '{$classname}' due to error: ...", DEBUG_DEVELOPER);
}
```

So a developer who forgets the attribute `#[identifier_attribute]` sees no error at all, the scope is just silently missing from the list.

docs/apis/subsystems/routing/scopes.md
andrewnicols: Thanks - I renamed it after I first wrote these docs /o\

docs/apis/subsystems/routing/defining-scopes.md
andrewnicols: Fixed

docs/apis/subsystems/routing/defining-scopes.md
andrewnicols: Technically it is true though - at resolution an exception will be thrown. The scope repository will catch it, but the openapi docs will not, and some other places may not either.

docs/apis/subsystems/routing/defining-scopes.md
```diff
@@ -0,0 +1,147 @@
...
+## Declaring the identifier {/* #declaring-the-identifier */}
+
+Every class in the hierarchy -- both the abstract grouping classes and the concrete leaf class -- must carry a `\core\router\scope\identifier_attribute` attribute, contributing the segment of the identifier for that level:
+
+```php
+#[\core\router\scope\identifier_attribute('content')]
+```
+
+The identifier passed to the attribute must start with a letter, and may otherwise only contain lowercase letters, numbers, and underscores. It must **not** contain a colon (`:`) or the component name -- both of these are added automatically.
```
meirzamoodle: Based on  the code:
```
// public/lib/classes/router/scope/identifier_attribute.php:46
if (!preg_match('/^[a-z][a-z0-9_]*$/', $identifier)) {
```
It said that the attribute must start with a lowercase letter. So, I think we should add it also to the doc: "The identifier passed to the attribute must start with a **lowercase** letter..."

Or we can drop "otherwise" and the sentence reads "must start with a letter and may only contain lowercase letters,..."

meirzamoodle (changes requested):

meirzamoodle (approved):

